PR #16, opened 2012-03-09
Hi Mike

beneiltis:
As I told you I added support for self signed certificates.
I used a CRT file created with java keytool and included it in the project bundle.
The file is hardcoded so you might wanne change it.

You can also adjust a lot of things (using the ios sec api) but I leave to you to decide what to use.
It is sufficient for our purpose but SocketRocket might demand more flexibility.

It would be great if this could become part of the official library.

Cheers 
Bene


dfed:

PR #40, opened 2012-06-15
Fix readme typeos and ignore more Xcode noise

soffes:


dfed:

PR #58, opened 2012-10-22
Added sendPing method

olegam:
Added sendPing method to test if the connection is still working. If the ping fails the connection will fail after the timeoutInterval specified in the initial NSURLRequest.


vigneshv59:
+1

dfed:

PR #73, opened 2013-01-21
Made the version string in the Podspec Ruby friendly

steveluscher:
Here's a minor change that keeps the pod linter from puking when you try to specify pre-release versions of SocketRocket in your Podfile. Use periods instead of hyphens!


dfed:

That said, our current release is not a `-beta`, but we'll keep that in mind going forward.

PR #74, opened 2013-01-31
change location of headers for iOS static lib target

pizthewiz:
loosen the requirement that an application/super-project must reference the subproject headers explicitly through the source tree; also avoids copying the headers at application archive time; see jamie montgomerie’s explanation here: http://www.blog.montgomerie.net/easy-xcode-static-library-subprojects-and-submodules

i should also mention, the PR should not require existing iOS SocketRocket clients to change how headers are referenced (as one could always use headers from the project source tree itself), but makes the headers available in a framework style #import <SocketRocket/SRWebSocket.h.h> much like one would find on Mac OS X after a header search path tweak.

### DIFF
--- a/SocketRocket.xcodeproj/project.pbxproj
+++ b/SocketRocket.xcodeproj/project.pbxproj
@@ -8,7 +8,6 @@
 
 /* Begin PBXBuildFile section */
 		F6016C7C146124B20037BB3D /* base64.c in Sources */ = {isa = PBXBuildFile; fileRef = F6016C7B146124B20037BB3D /* base64.c */; };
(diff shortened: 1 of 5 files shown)


SocketRocket.xcodeproj/project.pbxproj
```diff
@@ -705,6 +703,7 @@
 				);
 				OTHER_LDFLAGS = "-ObjC";
 				PRODUCT_NAME = "$(TARGET_NAME)";
+				PUBLIC_HEADERS_FOLDER_PATH = include/$PRODUCT_NAME;
```
efirestone: nit: The "standard" way to write this (as defined by those created by Xcode) is using parentheses: `$(PRODUCT_NAME)`. Not a blocker though.

efirestone:
Looks good after resolving conflicts.

dfed:
Closing this out since we duplicated this #292

PR #86, opened 2013-03-12
Decrement deployment target

### LINKED ISSUE #83: Targetting 10.6/10.7
I was able to successfully compile SocketRocket for 10.6 simply by changing the deployment target from 10.8 to 10.6. It's my understanding that ARCLite should take care of most of the core differences, unless the code relies on zeroing weak references.

Would you be interested in pull requests relevant to older OS X versions that may be easy to support? I'm using this in a project that doesn't necessarily need to work on 10.6 but am aiming for that unless we find a significant hurdle.

One such hurdle would be having to maintain a stale fork of a library that really prefers to support only newer versions ;-)

natevw:
This library works fine on Lion (and even compiles fine for 10.6 too, but haven't had a chance to actually test) and would appreciate the target being set lower. Potentially closes #83 unless we should just bump it down all the way to 10.6.


dfed:
Thanks for the PR @natevw. Sorry for taking so long to get back to you on this, but it looks good to me! Merging.

PR #88, opened 2013-03-27
Allow this framework to be bundled more flexibly on OS X

natevw:
(e.g. as a subframework within another bundle)

See these for more discussion of why this is better:
http://overooped.com/post/87887322/use-rpath-instead-of-loader-path-or-executable-path
http://www.dribin.org/dave/blog/archives/2009/11/15/rpath/

In my case, I needed SocketRocket to be a ± hidden dependency within a site-specific framework and the app could not be deployed with the previous installation path. (simply using `@loader_path` does work because within an app it would be at `@loader_path/../Frameworks` but within a parent framework it needs `@loader_path/Frameworks`.)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,6 +79,7 @@ more rhudimentary and less robust.
 2. Add ``SocketRocket.framework`` to the link libraries
 3. If you don't have a "copy files" step for ``Framework``, create one
 4. Add ``SocketRocket.framework`` to the "copy files" step.
+5. Add ``@loader_path/../Frameworks`` to the "Runpath Search Paths" setting of your app target
 
 API
 ---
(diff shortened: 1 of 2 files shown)


martinmroz:
Thanks for your pull request @natevw -- This looks like a worthwhile change and we'd like to take it. Looks like there's conflicts that have to be resolved before we merge the PR, and once that's taken care of we'll merge. Sorry for the delay! Promise it won't be another 2 years ^_^

kenwigginton:
@natevw We're doing a little repository cleaning so I'm going to close this out for now. Again, this looks like a worthwhile change and we would love to take it, but merge conflicts would need to be resolved first.

Thanks for your time!

PR #89, opened 2013-03-27
Fixed bad_access that could occur when releasing _workQueue on dealloc

janckoch:
Hi. 

I encountered a BAD_ACCESS on dealloc in SRWebSocket. It happened because you were not testing if _workQueue is nil before releasing it. This should fix it.

Cheers,
Jan


dfed:
Thanks for fixing! Merging.

PR #97, opened 2013-05-29
Fixed compiler warnings

graeme-hill:
With default xcode settings I had a bunch of warnings. Most of them were due to necessary implicit narrowing conversions so I made them explicit.


kenwigginton:
Hi @graeme-hill , thanks for your pull-request. These changes all look good, but looks like someone also fixed the same issues you were having outside of your request as well.  Socket-Rocket should be 64-bit compatible now. Closing as this is resolved as of now.

PR #98, opened 2013-05-30
fixed memory leak and added TSL workaround for iOS 5.0

adrianUken:
-renamed the block variable "self" to "webSocket" in several blocks to prevent confusion
-ensured references in blocks to self (and properties of self) properly reference the "webSocket" block variable to prevent retain loops
-added the TSL workaround for iOS 5.0 devices to properly connect on secure connections as described here https://developer.apple.com/library/ios/#technotes/tn2287/_index.html#//apple_ref/doc/uid/DTS40011309


shyreen:
This solution is not working on IOS 9.1

PR #109, opened 2013-08-14
Use the correct developer frameworks variable so that builds work properly under differing versions of Xcode

tonyarnold:


dfed:

PR #111, opened 2013-09-01
Add receivedHTTPHeaders array to public properties

aaronvegh:
This is my first-ever pull request; apologies if I'm not doing this right. 

For my application I need to get the actual headers served up by the web service in the event of a failure. However the receivedHTTPHeaders property is private in the current release. My change puts that property in the public so my app can access them. I'm not sure if there's a Very Good Reason™ why the headers aren't currently available, but this change was useful to me.

Thanks for reading! This has been such a terrific library of code to use, so thanks for all your great work.


SocketRocket.xcodeproj/project.xcworkspace/xcshareddata/SocketRocket.xccheckout
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dfed: This file should be in our .gitignore. We're adding that now. Please remove this file from your PR.

dfed:
This PR looks good. Merging now. @kenwigginton is going to remove the `SocketRocket.xccheckout` file in an upcoming PR.

PR #125, opened 2013-10-20
Update exported headers configuration #xcode5

mattdipasquale:
After following [Using Static Libraries in iOS:
Configuring Exported Headers][1], I created a new
Xcode project with the Cocoa Touch Static Library
template, compared the Copy Files build phase to
that of SocketRocket, and confirmed they were
identical.

I chose not to export `base64.h` because it's
undocumented (in the README).

  [1]: https://developer.apple.com/library/ios/technotes/iOSStaticLibraries/Articles/creating.html#//apple_ref/doc/uid/TP40012554-CH2-SW4


dfed:

PR #137, opened 2013-12-28
Allow the clients to decide if validation of SSL certificates should be skipped

jmkk:
Instead of just assuming that in DEBUG root certificates should not be evaluated allow the callers to decide.
In order to not break existing clients, in DEBUG evaluation still is off, but in release builds you can set the allowUntrustedSSLCertificates flag.


SocketRocket/SRWebSocket.h
```diff
@@ -41,6 +41,12 @@ extern NSString *const SRWebSocketErrorDomain;
 @property (nonatomic, readonly) SRReadyState readyState;
 @property (nonatomic, readonly, retain) NSURL *url;
 
+// Specifies whether SSL trust chain should NOT be evaluated.
+// By default this flag is set to NO, meaning only secure SSL connections are allowed.
+// For DEBUG builds this flag is ignored, and SSL connections are allowed regardless
+// of the certificate trust configuration
+@property (nonatomic, readwrite) BOOL allowUntrustedSSLCertificates;
```
martinmroz: Total nit, this should be allowsSelfSignedCertificates (https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/CodingGuidelines/Articles/NamingIvarsAndTypes.html) -- in the initializers it should be allows just like you've done.

More importantly though, to stop people from hurting themselves, this property should be marked readonly and the only way to specify it is in the initializer.

SocketRocket/SRWebSocket.m
```diff
@@ -581,21 +586,26 @@ - (void)_initializeStreams;
             [SSLOptions setValue:[NSNumber numberWithBool:NO] forKey:(__bridge id)kCFStreamSSLValidatesCertificateChain];
         }
         
+        BOOL allowUntrustedSSLCertificates = self.allowUntrustedSSLCertificates;
+        
 #if DEBUG
-        [SSLOptions setValue:[NSNumber numberWithBool:NO] forKey:(__bridge id)kCFStreamSSLValidatesCertificateChain];
```
martinmroz: Friendly note, might be easier to read `[NSNumber numberWithBool:NO]` if it was `@NO`!

SocketRocket/SRWebSocket.m
```diff
@@ -581,21 +586,26 @@ - (void)_initializeStreams;
...
-        NSLog(@"SocketRocket: In debug mode.  Allowing connection to any root cert");
+        allowUntrustedSSLCertificates = YES;
 #endif
+
+        if (allowUntrustedSSLCertificates) {
+            [SSLOptions setValue:[NSNumber numberWithBool:NO] forKey:(__bridge id)kCFStreamSSLValidatesCertificateChain];
```
martinmroz: Friendly note, might be easier to read `[NSNumber numberWithBool:NO]` if it was `@NO`! Wasn't you I know ^_^

martinmroz:
This looks great and we'd love to take this, just the last feedback on dependency injecting instead of having a mutable property! Thanks!

martinmroz:
Oh, there also appear to be some merge conflicts to resolve!

jmkk:
There @martinmroz, let's see if this is better. It's been a while! Thanks.

jmkk:
@martinmroz ...or do we want to wait another 2.5 years? ;-)

dfed:
It won't be that long, we promise! We're doing sweeps a minimum of monthly to try to clear out our backlog. Look for an answer this Friday 😃. Thanks for bearing with us!

kenwigginton:
Thank you @jmkk !

PR #142, opened 2014-01-13
allow user to send ping and listen to ping/pong

blackm00n:


dfed:

PR #148, opened 2014-02-04
Added cast to silence compiler warning

startupthekid:
Cast to (size_t) to silence warning about losing integer precision.


martinmroz:
Doing some house-cleaning and I believe this fix has already been taken from a different PR. Thanks!

PR #163, opened 2014-04-07
base64Encoding Deprecated

newshorts:
[Apple's documentation](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSData_Class/DeprecationAppendix/AppendixADeprecatedAPI.html#//apple_ref/occ/instm/NSData/base64Encoding)  states they are deprecating base64Encoding, updated the calls from base64Encoding to ```base64EncodedStringWithOptions:``` instead.

Not sure if you need a 64 character line length. Also, did not specify line endings so they will default to: Carriage Return + Line Feed.

Careful, this is only for OSX 10.9 +

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -124,7 +124,7 @@ static NSString *newSHA1String(const char *bytes, size_t length) {
     assert(length <= UINT32_MAX);
     CC_SHA1(bytes, (CC_LONG)length, md);
     
-    return [[NSData dataWithBytes:md length:CC_SHA1_DIGEST_LENGTH] base64Encoding];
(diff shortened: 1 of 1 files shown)


reqshark:
woah this PR is pretty old, but it might be getting more relevant

RoyalPineapple:
This PR has merge conflicts. Please rebase off master to fix.

dfed:
Also please fall back to `base64Encoding` if `base64EncodedStringWithOptions:` is unavailable. We should also just use `base64EncodedStringWithOptions:0`. Thanks for the PR!

martinmroz:
Thanks for your contributions! We've already addressed this particular issue by merging a different pull request.

PR #176, opened 2014-06-24
Adds sendPong for sending unsolicited pong messages.

WolfgangD:
According to the websocket specification a pong message can be sent unsolicited. This was not possible with the current API.

Pull requests adds new method sendPong for triggering an unsolicited pong message. Used and tested in an iOS app to keep the connection alive (heartbeat).


dfed:
Closing this out since our CLA isn't signed, and our comments haven't been addressed. That said, we would welcome another PR to send an unsolicited pong.

PR #191, opened 2014-08-19
Fix block retain cycles and rename local 'self' variables to avoid confusion

bgfriend0:
Several ivars were accessed in blocks, implicitly causing self to be retained. I verified that this caused a retain cycle on SRWebSocket, and these commits resolved the retain cycle for me.

Additionally, I renamed the local 'self' variables to avoid confusion with the 'self' instance pointer.

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -1425,9 +1425,6 @@ static const size_t SRFrameHeaderOverhead = 32;
 
 - (void)safeHandleEvent:(NSStreamEvent)eventCode stream:(NSStream *)aStream
 {
-    __strong typeof(self) strongSelf = self;
-    ((void)strongSelf);
-
     switch (eventCode) {
         case NSStreamEventOpenCompleted: {
             SRFastLog(@"NSStreamEventOpenCompleted %@", aStream);
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -248,6 +248,7 @@ typedef void (^data_callback)(SRWebSocket *webSocket,  NSData *data);
     
     BOOL _sentClose;
     BOOL _didFail;
+    BOOL _cleanupScheduled;
     int _closeCode;
     
     BOOL _isPumping;
@@ -670,7 +671,7 @@ static __strong NSData *CRLFCRLF;
             }];
 
             self.readyState = SR_CLOSED;
-            _selfRetain = nil;
+            [self _scheduleCleanup];
 
             SRFastLog(@"Failing with error %@", error.localizedDescription);
             
@@ -1083,7 +1084,7 @@ static const uint8_t SRPayloadLenMask   = 0x7F;
             }];
         }
         
-        _selfRetain = nil;
+        [self _scheduleCleanup];
     }
 }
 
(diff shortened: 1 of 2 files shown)


SocketRocket/SRWebSocket.m
```diff
@@ -110,8 +110,14 @@ @interface _SRRunLoopThread : NSThread
     assert(length >= 0);
     assert(length <= UINT32_MAX);
     CC_SHA1(bytes, (CC_LONG)length, md);
-    
-    return [[NSData dataWithBytes:md length:CC_SHA1_DIGEST_LENGTH] base64Encoding];
+
+    NSData *data = [NSData dataWithBytes:md length:CC_SHA1_DIGEST_LENGTH];
+    NSString *dataString;
+    if ([data respondsToSelector:@selector(base64EncodedStringWithOptions:)])
```
martinmroz: To be defensive, please use braces around the body of your `if` and `else` statements; after all, that's how [goto fail;](https://www.imperialviolet.org/2014/02/22/applebug.html) happened!

SocketRocket/SRWebSocket.m
```diff
@@ -499,7 +506,10 @@ - (void)didConnect
         
     NSMutableData *keyBytes = [[NSMutableData alloc] initWithLength:16];
     SecRandomCopyBytes(kSecRandomDefault, keyBytes.length, keyBytes.mutableBytes);
-    _secKey = keyBytes.base64Encoding;
+    if ([keyBytes respondsToSelector:@selector(base64EncodedStringWithOptions:)])
```
martinmroz: Please brace your `if` statements here to!

SocketRocket/SRWebSocket.m
```diff
@@ -477,14 +484,14 @@ - (void)_readHTTPHeader;
         _receivedHTTPHeaders = CFHTTPMessageCreateEmpty(NULL, NO);
     }
                         
-    [self _readUntilHeaderCompleteWithCallback:^(SRWebSocket *self,  NSData *data) {
-        CFHTTPMessageAppendBytes(_receivedHTTPHeaders, (const UInt8 *)data.bytes, data.length);
+    [self _readUntilHeaderCompleteWithCallback:^(SRWebSocket *webSocket,  NSData *data) {
```
martinmroz: Good catch here! Thanks!

SocketRocket/SRWebSocket.m
```diff
@@ -1378,90 +1417,98 @@ - (void)stream:(NSStream *)aStream handleEvent:(NSStreamEvent)eventCode;
...
-                [self _pumpWriting];
-                [self _pumpScanner];
-                break;
+        [weakSelf safeHandleEvent:eventCode stream:aStream];
+    });
+}
+
+- (void)safeHandleEvent:(NSStreamEvent)eventCode stream:(NSStream *)aStream
+{
+    __strong typeof(self) strongSelf = self;
```
martinmroz: Correct me if I'm wrong, but are you confident this does what you're intending? I believe in the context of a method invocation like this, `self` must be live (otherwise it would have short-circuited at `objc_msgSend(nil)`). Once in the context of an invocation like this, your reference to `self` would already be strong, and simply re-declaring it `__strong` explicitly won't actually do anything.

martinmroz:
Thanks for the PR! Please resolve merge conflicts and we'll take your fix.

dfed:
Ping on fixing the merge conflicts. We'd like to take this PR!

bgfriend0:
hey guys, i'll try to take a look at this soon. the PR is over a year old.

dfed:
Thanks! Sorry for the delay on our part.

SocketRocket/SRWebSocket.m
bgfriend0: I think you're 100% correct. As this PR was submitted over a year ago, I am having a hard time recalling the context, but this looks to me to be cruft that was left over from the modification.

bgfriend0:
@martinmroz @dfed hi guys, I made the line item changes you requested. bgfriend0/master is updated and this PR includes the latest. it is, however, conflicting with your current master, as a lot has happened in the last year. How do you want to resolve those conflicts? You are far more familiar with the modern intent.

dfed:
There's no easy way to deal with conflicts. Do your best to apply your changes to top of master?

nlutsenko:
@bgfriend0, since this is stale I believe this is no longer relevant.
Also the master diverged from the commit you are using quite dramatically, so you would need to merge it again :(

Please reopen/ping this PR if yo think it is relevant.

PR #196, opened 2014-08-28
VOIP Background Socket Support

thebehera:
Set the socket type to VOIP if it's set in the Info.plist file


lylepratt:
Would love to see this accepted

xzenon:
+1 to merge to enable support of VoIP apps.

PR #203, opened 2014-10-06
Keeping original NSURLRequest object HTTPMethod and HTTPBody

danielalves:
Not stripping the original `NSURLRequest` object `HTTPBody` and keeping its `HTTPMethod` gives us the possibility of doing some checks on the server before establishing the websocket. For example, you could have something like this using `ruby`, `sinatra` and `sinatra-websockets`:

```ruby
post '/web-socket-url' do
  if !request.websocket?
    return status 404
  end

  # Without HTTP method post and body data we can't do this in a proper way... Only if we sent
  # this parameter in the query string, what's not good for sensible data
  user_access_token = params[:user_token]

  if !( is_valid_token? user_access_token )
    return status 404
  end

  # Only now we establish the websocket
  request.websocket do |ws|

    ws.onopen do
      # ...
    end

    ws.onmessage do |message|
      # ...
    end

    ws.onclose do
      # ...
    end
  end
end
```

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -498,7 +498,12 @@ static __strong NSData *CRLFCRLF;
 - (void)didConnect
 {
     SRFastLog(@"Connected");
-    CFHTTPMessageRef request = CFHTTPMessageCreateRequest(NULL, CFSTR("GET"), (__bridge CFURLRef)_url, kCFHTTPVersion1_1);
+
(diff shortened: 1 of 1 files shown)


mikelikespie:
@danielalves  There's no reason to use anything but get request. The RFC states

> An HTTP/1.1 or higher GET request, including a "Request-URI" [RFC2616] that should be interpreted as a /resource name/ defined in Section 3 (or an absolute HTTP/HTTPS URI containing the /resource name/).

And that implies there is no HTTP body.

PR #207, opened 2014-10-25
Fix warnings

Autoc0diq:


dfed:
This has since been fixed elsewhere. Sorry for the delay, and thanks for the PR!

PR #212, opened 2014-11-21
Anchor certificates support

oleganza:
This patch allows using anchor certificates instead of pinned certificates if needed.


FredericJacobs:
Also needed this and implemented in #220.

dfed:
Closing this in favor of #220.

PR #214, opened 2015-01-05
Ignore deprecation warnings for older iOS versions

pnc:
The library falls back to the `-base64Encoding` to support iOS versions older than 7.0, but this triggers a build warning. This change silences the build warning, since these selectors are only used when the replacement isn't available.


natan:
@pnc Thanks! Can you please sign our [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1)?

pnc:
@natan Done!

PR #220, opened 2015-01-30
Evaluate certificate chain trust using SecTrustEvaluate

FredericJacobs:
The current pinning approach is quite restrictive. This PR fixes the issue of pinning with signing certificates instead of direct public key pinning.
This should work for current users without changing their code.
Uses the built-in chain evaluation function.

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -1389,7 +1389,7 @@ static BOOL evaluateServerTrust(SecTrustRef serverTrust) {
 
     OSStatus status = SecTrustEvaluate(serverTrust, &result);
     
-    isValid = ((result == kSecTrustResultUnspecified || result == kSecTrustResultProceed) && status == 0);
+    isValid = ((result == kSecTrustResultUnspecified) && status == 0);
 
     return isValid;
 }
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -1383,6 +1383,16 @@ static const size_t SRFrameHeaderOverhead = 32;
     
     [self _writeData:frame];
 }
+static BOOL evaluateServerTrust(SecTrustRef serverTrust) {
+    BOOL isValid = NO;
+    SecTrustResultType result;
+
+    OSStatus status = SecTrustEvaluate(serverTrust, &result);
+    
+    isValid = ((result == kSecTrustResultUnspecified || result == kSecTrustResultProceed) && status == 0);
+
+    return isValid;
+}
 
 - (void)stream:(NSStream *)aStream handleEvent:(NSStreamEvent)eventCode;
 {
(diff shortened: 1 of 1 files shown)


FredericJacobs:
Thoughts @natan, @alanf?

SocketRocket/SRWebSocket.m
```diff
@@ -1383,6 +1383,16 @@ - (void)_sendFrameWithOpcode:(SROpCode)opcode data:(id)data;
     
     [self _writeData:frame];
 }
+static BOOL evaluateServerTrust(SecTrustRef serverTrust) {
+    BOOL isValid = NO;
+    SecTrustResultType result;
+
+    OSStatus status = SecTrustEvaluate(serverTrust, &result);
+    
+    isValid = ((result == kSecTrustResultUnspecified || result == kSecTrustResultProceed) && status == 0);
```
emonti: If you are pinning, you may actually only want to consider kSecTrustResultUnspecfied valid.

SecTrustResultProceed is defined as "The user indicated that you may trust the certificate for the purposes designated in the specified policies. ". FWIW, though, these constants are confusingly named by Apple.

See: https://developer.apple.com/library/mac/documentation/Security/Reference/certifkeytrustservices/index.html#//apple_ref/c/tdef/SecTrustResultType

SocketRocket/SRWebSocket.m
```diff
@@ -1390,26 +1400,10 @@ - (void)stream:(NSStream *)aStream handleEvent:(NSStreamEvent)eventCode;
...
-                        
-                        if ([trustedCertData isEqualToData:certData]) {
-                            _pinnedCertFound = YES;
-                            break;
-                        }
-                    }
-                }
-            }
+            SecTrustRef serverTrust = (__bridge SecTrustRef)[aStream propertyForKey:(__bridge id)kCFStreamPropertySSLPeerTrust];
+            SecTrustSetAnchorCertificates(serverTrust, (__bridge CFArrayRef)sslCerts);
```
emonti: I think you also want to ignore all other root certificates with "SecTrustSetAnchorCertificatesOnly"

See:
https://developer.apple.com/library/mac/documentation/Security/Reference/certifkeytrustservices/index.html#//apple_ref/c/func/SecTrustSetAnchorCertificatesOnly

SocketRocket/SRWebSocket.m
FredericJacobs: If you're pinning a certificate that is in the user's trust store wouldn't that return a `SecTrustResultProceed`?

SocketRocket/SRWebSocket.m
dfed: @emonti is this a merge blocker? If not, this looks good to me.

FredericJacobs:
This PR should be closed and replaced by #274

PR #225, opened 2015-02-21
Added "Carthage" package manager support

dodikk:
* Added iOS dynamic framework target
* Added iOS dynamic framework shared scheme
* Added a protocol for mocking the web socket in user's unit tests
* Introduced more strict warning level

PR #227, opened 2015-02-25
Replaced "extern NSString* const" with class

dodikk:
Attempting to resolve a linker conflict with Segment.IO
https://github.com/segmentio/analytics-ios/issues/258

Includes [carthage support](https://github.com/square/SocketRocket/pull/225) patches.


SocketRocket-iOSTests/SocketRocket_iOSTests.m
```diff
@@ -0,0 +1,40 @@
+//
```
martinmroz: This file appears to be just boilerplate; consider removing it as noise.

dfed:
There's a lot going on here. Let's separate out individual features with individual PRs. At Square we have a principal to keep PRs under 500 lines for easy reading. Closing this in the meantime. Thanks!

PR #230, opened 2015-03-18
Read network service type from NSURLRequest instances

jessearmand:
In order to use SocketRocket with background mode, we have to set the NSURLNetworkServiceType on the NSURLRequest instance passed to it. 

But currently, these network services types are not applied to the input and output streams being used by SocketRocket. This pull request is adding that support.

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -27,6 +27,7 @@
 
 #if TARGET_OS_IPHONE
 #import <Endian.h>
+#import <UIKit/UIKit.h>
 #else
 #import <CoreServices/CoreServices.h>
 #endif
@@ -592,12 +593,14 @@ static __strong NSData *CRLFCRLF;
         case NSURLNetworkServiceTypeDefault:
             return;
         case NSURLNetworkServiceTypeVoIP: {
+#if TARGET_OS_IPHONE
             // voice over IP control - this service type is deprecated in favor of using PushKit for VoIP control
-            if (floor(NSFoundationVersionNumber) > NSFoundationVersionNumber_iOS_8_3) {
-                return;
+            if ([[UIDevice currentDevice].systemVersion compare:@"9.0" options:NSNumericSearch] == NSOrderedAscending) {
+                networkServiceType = NSStreamNetworkServiceTypeVoIP;
+                break;
             }
-            networkServiceType = NSStreamNetworkServiceTypeVoIP;
-            break;
+#endif
+            return;
         }
         case NSURLNetworkServiceTypeVideo:
             networkServiceType = NSStreamNetworkServiceTypeVideo;
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -586,6 +586,26 @@ static __strong NSData *CRLFCRLF;
     
     _inputStream.delegate = self;
     _outputStream.delegate = self;
+    
+    NSString *networkServiceType;
+    switch (_urlRequest.networkServiceType) {
+        case NSURLNetworkServiceTypeVoIP:
+            networkServiceType = NSStreamNetworkServiceTypeVoIP;
+            break;
+        case NSURLNetworkServiceTypeVideo:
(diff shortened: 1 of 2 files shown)


SocketRocket/SRWebSocket.m
```diff
@@ -586,6 +586,26 @@ - (void)_initializeStreams;
     
     _inputStream.delegate = self;
     _outputStream.delegate = self;
+    
+    NSString *networkServiceType;
+    switch (_urlRequest.networkServiceType) {
+        case NSURLNetworkServiceTypeVoIP:
+            networkServiceType = NSStreamNetworkServiceTypeVoIP;
```
RoyalPineapple: ServiceTypeVoIP is deprecated in iOS 9, please handle.

SocketRocket/SRWebSocket.m
```diff
@@ -586,6 +586,26 @@ - (void)_initializeStreams;
...
+        case NSURLNetworkServiceTypeVideo:
+            networkServiceType = NSStreamNetworkServiceTypeVideo;
+            break;
+        case NSURLNetworkServiceTypeBackground:
+            networkServiceType = NSStreamNetworkServiceTypeBackground;
+            break;
+        case NSURLNetworkServiceTypeVoice:
+            networkServiceType = NSStreamNetworkServiceTypeVoice;
+            break;
+        default:
```
mikelikespie: Can we levae the default case off so we get a warning when new networkServiceTypes are added?

RoyalPineapple: Please remove the default case, handle all cases.

mikelikespie:
@jessearmand After that one minor change, could you sign our [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1) as outlined in our [README here](https://github.com/square/SocketRocket#contributing)?

jessearmand:
@mikelikespie I have signed the CLA, but the build will fail if the Travis CI is not set to:

    osx_image: xcode7

As described here: http://blog.travis-ci.com/2015-09-09-xcode7-gm/

NSFoundationVersionNumber_iOS_8_3 is only available on ios 9.0, and I'm not sure if using: 

    NSString *osVersion = [[UIDevice currentDevice] systemVersion];

is a better way to check compatibility.

dfed:
Using `[[UIDevice currentDevice] systemVersion];` is backwards compatible, and therefore should be used. We shouldn't need to bump Xcode to land this PR. Thanks!

jessearmand:
@dfed I think the build and test setup is getting more and more complicated because I need to import UIKit because of this UIDevice requirement, will get back to you on this when I have the time.

dfed:
Maybe UiDevice isn't the right way to go then if we aren't already importing UIKit. We can check if particular pointers exist as a fallback – check out [what Valet does to determine if the app is being run on a particular OS](https://github.com/square/Valet/blob/master/Valet/VALSecureEnclaveValet.m#L36). Whatever we do, it should be backwards compatible.

jessearmand:
I made another pull request #293. Also I've checked how Valet does it. But, I think it needs an additional check at runtime to determines the actual OS that it's running on. That's where I'm using the `NSFoundationVersionNumber` comparison. 

Because if it's build on Xcode that has `__IPHONE_9_0`, but targeted at iOS 8 devices, then it will still execute that specific part of code, am I right?

dfed:
You're right @jessearmand. Once Valet checks to see if you're on the right build system (`VAL_IOS_8_OR_LATER`), it checks to see if the pointer value is nil for constants defined in iOS 9 but don't exist in iOS 8 (`&kSecAttrAccessControl != NULL`). You could use a similar approach in #293, but it looks like you've got it building, so all is well :)

jessearmand:
Oh I see, I wasn't clear about checking pointers being NULL. But, in my pull request, it's about handling deprecation,`kCFStreamNetworkServiceTypeVoIP` is available from iOS 4.0 and later.

dfed:
yes indeed. the pointer value check is just a nifty (if slightly obtuse) way to tell what OS version you're on. Your way works too!

PR #231, opened 2015-03-18
Fix issue where we make the GET before checking the Server-side cert

seanwhitsell:
Pentesting on our app shows that when we choose to pin to a certificate And the Web Socket Server requires an Authorization: header, then SocketRocket will make the GET (including the Authorization: header) before testing the Server Certificate.

This change moves the sending of the GET to a place in the NSStream work flow where we have the Server Cert to test against.

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -1415,6 +1415,12 @@ static const size_t SRFrameHeaderOverhead = 32;
                 });
                 return;
             }
+            
+            if (aStream == _outputStream && _pinnedCertFound) {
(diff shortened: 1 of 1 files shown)


dfed:
Hi @seanwhitsell, thanks for the PR. This looks good to me. Only blocker is you need to sign our [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1). Let us know when you've signed, and we'll merge.

seanwhitsell:
Thanks Dan.

The CLA has been signed.

Cheers,

Sean

Issue #232, opened 2015-03-18
XCode 6.3 beta warnings about extra semicolons in SRWebSocket.m 


jackpal:
When I compile SRWebSocket.m using XCode 6.3 I get numerous warnings of this form:

SRWebSocket.m:128:45: Semicolon before method body is ignored
SRWebSocket.m:138:45: Semicolon before method body is ignored

dfed:
This issue has been resolved by Xcode fixing itself :)

PR #233, opened 2015-03-19
Remove semicolons before method bodies.

jackpal:
Fixes XCode 6.3 beta 2 warnings.

Fixes issue #232


dfed:
This doesn't cause warnings on Xcode 6.3 or 6.4. I don't think we need this change. Sounds like 6.3 Beta 2 had a bug.

PR #234, opened 2015-04-08
Fix #226: AppStore rejection when using _connect

### LINKED ISSUE #226: App store rejection
The use of _connect method in SRWebSocket.m is causing the app being rejected in the automated screening process during an app store submission.

Here's a reply from the Apple review team.
> This issue can occur if an app calls names that are similar to or directly calls our non-public APIs. It would be appropriate to remove all references to the selectors listed in the error message you received before resubmitting this app for review. I have verified that _connect shares a name with a private API and that it falls into this category of selectors.

frobichaud:


dfed:
Fixing this with #252. thanks for bringing this to our attention!

PR #239, opened 2015-04-29
General tweaks: performance/safety/API/documentation

noremac:
Hello, SocketRocket is great! I had one problem with it though relating to my need to send large files. I wanted an API where I could send the socket some partial data along with a write identifier and then receive a callback with the write identifier when it was done so I could queue up the next piece of data to write.

I've added a new method on SRWebSocket to do just that:
```- (void)sendPartialData:(NSData *)message withIdentifier:(id)identifier;```

Upon write completion you would receive the following callback:
```- (void)webSocket:(SRWebSocket *)webSocket writeDidFinishWithIdentifier:(id)identifier;```

While I was making these changes I noticed a few TODOs relating to performance that I've addressed, and I scratched a few other minor itches of mine too. I'll document each of the changes below and I'd love to have my request considered for inclusion. Please let me know if you'd like me to modify/clarify/break up/change anything! Thanks again for providing SocketRocket to the community.

## Security
I noticed that ```DEBUG``` mode would always allow insecure SSL connections. I occasionally need this behavior when in ```Release``` mode as well, so I added a new property: ```allowInsecureConnections``` which defaults to ```YES``` in ```DEBUG``` and ```NO``` in ```Release```. This way, the old behavior is preserved, a little flexibility has been added, and the default behavior in ```Release``` mode is still safe.

## Performance
I noticed a few ```TODOs``` about performance while looking through the code that I tried my hand at fixing.

* In the ```send:``` method, the data was being copied in every time. I went ahead and added a property ```sendDataSafely``` which defaults to YES. This will continue to copy everything in to the ```send:``` method. When set to ```NO``` these extra copies are not performed. The client can safely set this to ```NO``` if they know they are always sending immutable objects. I was thinking we could just check the type, so if it was an ```NSMutableString```, or ```NSMutableData```, that we could do the copy, but that still won't be safe for ```NSData``` that was initialized with one of the ```noCopy``` initializers which might not guarantee it's internal buffers is still around by the time we actually use it. Either way, the default behavior is safe, and those that understand the implications can gain a small amount of performance by flipping a switch.

* There was an area in the ```_innerPumpScanner``` with a comment about superfluous NSData copies that I believe I've addressed as well. My change is on line 1324 for reviewal.

## General

* I remade the SRReadyState enum names to play more nicely in Swift. The old values still exist as well for backwards compatibility.
* I cleaned up many trivial warnings that Xcode 6.3 was complaining about. Things like extra semicolons after method declarations in the implementation file.
* I added two other new write methods, ```sendString:``` and ```sendData:``` for all those type-safety nuts out there :-)
* I added Apple Doc style docs to every method in the header. My grammar isn't the best, but I gave it a shot anyway.

Thanks again for your consideration.


dfed:
Hi @noremac. Than's for the PR! There's a lot going on here; more than we'd be willing to merge. Any chance you could break these individual changes out into separate PRs? Say, one for Documentation, one for Performance, and one for Security.

That said, renames that break backwards compatibility will not be merged. Whatever features are added, they should be backwards compatible.

PR #242, opened 2015-05-08
Update podspec

cbrauchli:
Would be great to make versions newer than two years old accessible via CocoaPods. I'm happy to use different version number if you prefer, or reference a commit directly, but I think it make sense to tag and push this version. Let me know!

### DIFF
--- a/SocketRocket.podspec
+++ b/SocketRocket.podspec
@@ -1,11 +1,11 @@
 Pod::Spec.new do |s|
   s.name               = "SocketRocket"
-  s.version            = '0.3.1-beta2'
+  s.version            = '0.3.2'
   s.summary            = 'A conforming WebSocket (RFC 6455) client library.'
   s.homepage           = 'https://github.com/square/SocketRocket'
   s.authors            = 'Square'
   s.license            = 'Apache License, Version 2.0'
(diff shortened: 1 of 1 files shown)


cbrauchli:
@robertcrabtree, if the owners tag that commit, using the version works fine too. And also has the benefit of showing up in the "Releases" section of the github project.

I switched to using the commit though, just because I would like this to be as easy to merge as possible.

GorkaMM:
+1
Yes please, update the Podspec.

saba-okkhq:
+1
Thank you.

robertcrabtree:
@cbrauchli Gotcha. BTW, did you run 'pod spec lint' on this? I had to fix one warning in the sources (typecast from NSUInteger -> int32_t), and I had to make additional modifications to the pod spec to get 'pod spec lint' to pass (eg. remove -Wno-format).

CocoaPods have stricter requirements now.

You can reference the commits I made to my fork if you like: https://github.com/robertcrabtree/SocketRocket/commits/0.3.2

dnkoutso:
Looks like this is obviated by #243 which now points to a tag instead of a commit sha instead.

PR #244, opened 2015-05-28
Fix 64 bit build error

james-howard:
SRWebSocket.m:1696:32: error: implicit conversion loses integer precision: 'unsigned long' to 'int32_t' (aka 'int') [-Werror,-Wshorten-64-to-32]
            return data.length - i;
            ~~~~~~ ~~~~~~~~~~~~^~~


dfed:
LGTM cc @justinseanmartin

justinseanmartin:
LGTM

PR #258, opened 2015-07-12
recommend websocketd

reqshark:
just discovered SocketRocket today. :+1:  

only want to suggest a link to [`websocketd`](https://github.com/joewalnes/websocketd) since i saw talk of dead simple websockets.


martinmroz:
Hi @reqshark -- thank you for your pull request! Unfortunately, we haven't used `websocketd` ourselves so we don't feel comfortable making this kind of recommendation.

PR #261, opened 2015-07-17
Make SRWebSocket.h public

dnkoutso:
The `SocketRocketOSX` target has it as public.
The `.podspec` specifies all source files including the `.h` file. It does not specify `public_header_files` which means by default it is treated as a public header by CocoaPods.

Any reason why it was not public before?

/cc @mikelikespie @dfed


dfed:
No reason that I know of, but @mikelikespie seems like the best person to answer it. In any case, GORM.

mikelikespie:
LGTM

Issue #265, opened 2015-07-30
Proxy support with SSL


katowulf:
For us, #114 is still relevant. Although, in our case, we only support encrypted traffic, so we would need this to work with HTTPS traffic. Firebase is using SocketRocket for our iOS SDK and currently, customers are affected by lack of proxy support. This is fairly common with corporate policy in the education sector. 

If SocketRocket could allow proxies with SSL it would allow us to push out a new release to resolve this issue.

peterfdo:
@mikelikespie , Agree with @katowulf on #114. Almost all of our education users in Australia and New Zealand, and many other international users, are *blocked completely* from using our app running Firebase. We have *no workaround* to address the issue. Our app is designed for use in schools, and given that it's summer here in the US, the majority of our traffic is from those southern hemisphere countries right now. We understand that you have other priorities and appreciate any help we can get, but until then, our active user base is dwindling because of this seasonal traffic.

angrykilobyte:
Also agree with @katowulf on #114. Our application runs in a corporate environment, and many customers would like to have their users' devices go through their proxy.

shofmn:
Agree with the posts above. I'm working for a worldwide software company (top 5) and we also experience this issue that we can't open a websocket secure connection within the corporate network due to missing proxy support.
Would be great to have this enabled in SocketRocket - then this would definitely be the library of our choice!

dfed:
Reopened the original. Thanks for letting us know! Closing this.

PR #271, opened 2015-09-08
Require TLS 1.2 & enable pinning.

FredericJacobs:
- Enforce TLS 1.2 connections for secure connections. Prevents downgrade attacks
- I think a lot of applications are using this library along with `AFNetworking` or another HTTP library, therefore I abstracted out the pinning so that they can write their own pinning logic according to their needs. It can also help issues like #265 where they might have to allow proxying certificates or others.
- `if ([trustedCertData isEqualToData:certData])` is a very naïve way to do pinning. Certificates might be re-issued and the expiration dates or validation of the domain is done nowhere. It shouldn't matter too much if certificates are pinned but implementers would enjoy getting some more flexibility there.

### DIFF
--- a/SocketRocket/SRWebSocket.h
+++ b/SocketRocket/SRWebSocket.h
@@ -17,6 +17,13 @@
 #import <Foundation/Foundation.h>
 #import <Security/SecCertificate.h>
 
+@protocol CertificateVerifier <NSObject>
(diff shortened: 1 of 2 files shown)


FredericJacobs:
Missing some methods

Issue #95, opened 2013-05-16
VOIP and background sockets


robreuss:
A common use case for WebSockets is to enable a VOIP app to receive call notifications while in the background.

The issues involved in keeping a socket alive in the background for a VOIP app are discussed in the programming guide on page 121: 

http://developer.apple.com/library/ios/documentation/iphone/conceptual/iphoneosprogrammingguide/iphoneappprogrammingguide.pdf

It seems like SocketRocket uses NSStream, so supporting this functionality would involve optionally setting a property on the stream where NSStreamNetworkServiceType is set to NSStreamNetworkServiceTypeVoIP.

Any thought to adding this as a configuration option?

FYI, my specific requirement is to support this through libPusher, which would in the best case leverage that parameter to expose it's own version:

https://github.com/lukeredpath/libPusher/issues/96

lylepratt:
We are experiencing this deficiency now as well. Is there a good reason the PR has not been accepted?

mikelikespie:
can the PR be reopened? Also need CLA to be signed.

MihaMajcen:
I think that correct way for implementing this feature is to set networkServiceType in NSURLRequest which is passed into init function. This way we can support multiple serviceTypes not just voip. 

I've implemented it in #312

nlutsenko:
Closing as it is indeed fully implemented via `NSURLRequest` `networkServiceType` property and is available in the latest release.

PR #275, opened 2015-09-11
Set voip property on streams with respect of UIBackgroundModes

darvin:


dfed:
Hi @darvin! Thanks for the PR. To contribute to SocketRocket you must first sign our [CLA](https://docs.google.com/a/squareup.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1).

SocketRocket/SRWebSocket.m
```diff
@@ -534,6 +534,14 @@ - (void)_initializeStreams;
     
     CFStreamCreatePairWithSocketToHost(NULL, (__bridge CFStringRef)host, port, &readStream, &writeStream);
     
+    //If voip is set in UIBackgroundMode, set apropriate properties on read and write stream
+    NSArray* backgroundModes = [[[NSBundle mainBundle] infoDictionary] objectForKey:@"UIBackgroundModes"];
```
dfed: Rather than detecting background modes manually and hard-coding iOS configuration strings into the project, it's probably better to use a Dependency Injection model and create a constructor that takes an array of service types to set for reading and writing (possibly with different parameters for the read and write stream). But I'll defer to @mikelikespie on the best way to get this functionality in here.

SocketRocket/SRWebSocket.m
```diff
@@ -534,6 +534,14 @@ - (void)_initializeStreams;
...
+    
+    if (backgroundModes && [backgroundModes containsObject:@"voip"]) {
+        CFReadStreamSetProperty(readStream, kCFStreamNetworkServiceType, kCFStreamNetworkServiceTypeVoIP);
```
dfed: `kCFStreamNetworkServiceTypeVoIP` has been deprecated in iOS 9, which might make this PR a non-starter. Again, I'll defer to @mikelikespie

darvin:
@dfed CLA is signed

SocketRocket/SRWebSocket.m
darvin: well, it's necessary for ios 8.0.

darvin:
Seems to be a popular issue:

https://github.com/lukeredpath/libPusher/issues/96
https://github.com/square/SocketRocket/issues/95
http://stackoverflow.com/questions/16883295/unable-to-keep-websocket-alive-in-background-ios

dfed:
Thanks for signing! I don't doubt that the issue is popular. I'm advocating for a fix that isn't a one-off, doesn't hardcode strings that might change, and is forward compatible with iOS 9.

SocketRocket/SRWebSocket.m
mikelikespie: Agreed with @dfed on this. We should probably add a category to NSURLRequest using the whole `+[NSURLProtocol propertyForKey:inRequest:]` set of functions to control this.

SocketRocket/SRWebSocket.m
mikelikespie: I agree that we shouldn't add new code that depends on a deprecated API. I'd be OK if we extracted this to a function that used the new behavior with PushKit if available and fell back to using deprecated API if not available.

darvin:
@dfed so thats the line that is questionable?
```
if (backgroundModes && [backgroundModes containsObject:@"voip"]) {
```

?

dfed:
The problems are:
1) You're opting in all streams for any app that has "voip" as a background mode – utilizing VOIP should be a manual opt in (@mikelikespie suggested a category)
2) You're utilizing deprecated API – this PR fix should utilize the current API, and fall back to deprecated API when using older system versions.
3) You're using hardcoded strings (i.e. "voip" and "UIBackgroundModes"), though fixing 1 will obviate this issue.

dfed:
Closing this for now given the above issues. Once they have been addressed, please reopen and link this PR.

PR #277, opened 2015-09-19
Update gitconfig

kenwigginton:
Update the .gitconfig to prevent xccheckout and other user-specific settings.


dfed:
yolo
LET'S DO IT

PR #278, opened 2015-09-19
Update gitignore 2

kenwigginton:


mikelikespie:
LG

kenwigginton:
The previous attempt removed some python gitignores. This reverts to the previous git ignore with less fanciness but is more concise.

PR #133, opened 2013-12-23
Added ability to provide cookies for the web socket connection

jmkk:
For web sockets used behind an already authenticated HTTP(S) session you might want to provide the session cookie or other cookies appropriate for the web socket connection (for example Spring's SessionID when connecting to a Spring-authentication java backend server).

Sample usage for the clients could be:

```
[[self webSocket] setRequestCookies:[[NSHTTPCookieStorage sharedHTTPCookieStorage] cookies]];
```

Note that you might want to limit the cookies provided to the WebSocket based on the URL (cookiesForURL:).

PR #280, opened 2015-09-19
Added ability to provide cookies for the web socket connection

jmkk:
For web sockets used behind an already authenticated HTTP(S) session you might want to provide the session cookie or other cookies appropriate for the web socket connection (for example Spring's SessionID when connecting to a Spring-authentication java backend server).

Sample usage for the clients could be:

```
[[self webSocket] setRequestCookies:[[NSHTTPCookieStorage sharedHTTPCookieStorage] cookies]];
```

Note that you might want to limit the cookies provided to the WebSocket based on the URL (cookiesForURL:).

This is a refreshed PR from https://github.com/square/SocketRocket/pull/133


dfed:
Thanks for re-opening! Looks great

PR #279, opened 2015-09-19
Added tvOS to podspec

vigneshv59:
Added the tvOS platform to the podspec.

PR #282, opened 2015-09-21
Add tvos as deployment target

hamchapman:
As the title says, add support for tvOS as a deployment target. 

I've tested it here: https://github.com/hamchapman/libPusherTVOSTest with an app that uses libPusher (https://github.com/lukeredpath/libPusher) which in turn makes use SocketRocket and everything seemed to be working fine.

Still waiting for Cocoapods to merge in the tvOS stuff as well (https://github.com/CocoaPods/Core/pull/263)


dfed:
This looks like a duplicate of #279

PR #293, opened 2015-10-06
Read network service types of NSURLRequest.

jessearmand:
In relation to pull request #230.

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -608,6 +608,46 @@ static __strong NSData *CRLFCRLF;
         [_outputStream setProperty:SSLOptions
                             forKey:(__bridge id)kCFStreamPropertySSLSettings];
     }
+    
+    _inputStream.delegate = self;
+    _outputStream.delegate = self;
+    
+    [self setupNetworkServiceType:_urlRequest.networkServiceType];
+}
+
+- (void)setupNetworkServiceType:(NSURLRequestNetworkServiceType)requestNetworkServiceType
+{
+    NSString *networkServiceType;
+    switch (requestNetworkServiceType) {
+        case NSURLNetworkServiceTypeDefault:
+            break;
+        case NSURLNetworkServiceTypeVoIP: {
+            networkServiceType = NSStreamNetworkServiceTypeVoIP;
+#ifdef __IPHONE_9_0
+            if (floor(NSFoundationVersionNumber) > NSFoundationVersionNumber_iOS_8_3) {
+                static dispatch_once_t predicate;
+                dispatch_once(&predicate, ^{
+                    NSLog(@"SocketRocket: %@ - this service type is deprecated in favor of using PushKit for VoIP control", networkServiceType);
+                });
+            }
+#endif
+            break;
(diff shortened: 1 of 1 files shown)


dfed:
That's for reopening! This looks good to me. Do we know if it can have any undesired effects on existing clients? i.e. should we make users opt-in to getting this flag? Or is this the expected behavior?

cc @mikelikespie

SocketRocket/SRWebSocket.m
```diff
@@ -570,6 +570,33 @@ - (void)_initializeStreams;
...
     _outputStream.delegate = self;
+    
+    NSString *networkServiceType;
+    switch (_urlRequest.networkServiceType) {
+        case NSURLNetworkServiceTypeDefault:
+            return;
+        case NSURLNetworkServiceTypeVoIP: {
+#ifdef __IPHONE_9_0
+            // voice over IP control - this service type is deprecated in favor of using PushKit for VoIP control
+            if (floor(NSFoundationVersionNumber) > NSFoundationVersionNumber_iOS_8_3) {
```
mikelikespie: This seems like it would create unexpected behavior if one is on a higher version and expects it to be a VoIP service (even if they ignore deprecation warnings).

What if we just logged a warning inside a dispatch_once saying its deprecated, but we're going to do it anyways?

SocketRocket/SRWebSocket.m
```diff
@@ -570,6 +570,33 @@ - (void)_initializeStreams;
     
     _inputStream.delegate = self;
     _outputStream.delegate = self;
+    
+    NSString *networkServiceType;
+    switch (_urlRequest.networkServiceType) {
+        case NSURLNetworkServiceTypeDefault:
+            return;
```
mikelikespie: I'd rather us not early `return` out of this. A break here seems like it would be fine. Having returns here could lead to undesired behaviors if somebody adds initialization logic to the end of this function.

My recommendation would be to move this into its own function called `initializeNetworkServiceType:` and inside that you can use early returns.

SocketRocket/SRWebSocket.m
jessearmand: I'm actually still in the process of examining this. In my app, I'm not sure if `PushKit` is enough to ensure that websocket data are continuously streamed after push notification is received in the background.

Yes, I'm also thinking about using a warning.

SocketRocket/SRWebSocket.m
jessearmand: A separate method could be better. 

I'm not sure about the safety of setting `networkServiceType` if it's break here. Because if the case is `NSURLNetworkServiceTypeDefault` it shouldn't set anything on the input & output stream.

mikelikespie:
LGTM. Thanks for working with us on this, @jessearmand. We'll merge this once conflicts are resolved.

jessearmand:
I encountered an issue with #231 where the websocket keep on closing its connection. 

I haven't done a thorough inspection / debugging on both ios and the websocket server. So, I am basing my patch on top of ead962232d93be42a0f82539d8aa22aaa931942b.

PR #294, opened 2015-10-08
Prevent SocketRocket killing the connection before notifying of final messages

joenoon:
We are noticing a race condition when connecting to Firebase from our React Native project, which itself uses a fork of SocketRocket.  I believe the code is the same in the place in question, so I wanted to see if we could get this verified/fixed in SocketRocket, and then get React Native to pick up the change if valid.

The issue appears to be a race between hitting NSStreamEventEndEncountered and receiving any final messages the server may have sent before closing the connection.

My patch just changes closing/cleanup from running immediately, to instead happen on the next dispatch_async _workQueue.  I haven't been able to reproduce the issue since making this change.  _failWithError itself already shells out to dispatch_async _workQueue, so this patch would seem consistent with that behavior.  But we really need someone who knows the internals/implications of this to take a look.

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -1435,20 +1435,22 @@ static const size_t SRFrameHeaderOverhead = 32;
                 if (aStream.streamError) {
                     [self _failWithError:aStream.streamError];
                 } else {
-                    if (self.readyState != SR_CLOSED) {
-                        self.readyState = SR_CLOSED;
-                        _selfRetain = nil;
-                    }
+                    dispatch_async(_workQueue, ^{
+                        if (self.readyState != SR_CLOSED) {
+                            self.readyState = SR_CLOSED;
+                            _selfRetain = nil;
+                        }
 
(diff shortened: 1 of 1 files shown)


dfed:
Hi @joenoon! Thanks for the PR. Looks good to me, but we can't merge until you've signed our [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1) (per our [README](https://github.com/square/SocketRocket/#contributing)).

@mikelikespie can you take a quick look and make sure I'm not missing anything? Thanks!

joenoon:
@dfed thanks, just signed

joenoon:
Any thoughts on this?  We're running successfully with this patch, but if we remove it, we run into the race conditions mentioned above.

dfed:
cc @mikelikespie 

You'll hear from us by Friday at the latest.

mikelikespie:
Thanks!

PR #304, opened 2015-10-29
Support public key pinning

tang3w:
This pull request implement the public key SSL pinning for SocketRocket.


tang3w:
It seems that the Xcode version for CI is less than 7.0?

The added code depends on macros in Xcode 7. Would you please upgrade Xcode 7.0 or later?

dfed:
Our CI version is set to older versions since we don't want to break them. That said, we're currently looking into adopting pinning with #274

PR #306, opened 2015-11-10
Namespace category names to avoid collisions

jder:
I encountered a warning in Xcode 7.1 when a third-party had pulled in their own copy of SRWebsocket & failed to namespace these two categories (they just replaced SR with their own initials, so missed these). I'm not sure if there were runtime consequences (the docs say these names must be unique but it's not clear what happens if they're not). 

Since everything else in SocketRocket is nicely namespaced, I assumed that this was just an oversight and you'd like to namespace them.

### DIFF
--- a/SocketRocket/SRWebSocket.h
+++ b/SocketRocket/SRWebSocket.h
@@ -115,17 +115,17 @@ extern NSString *const SRHTTPResponseErrorKey;
 
 @end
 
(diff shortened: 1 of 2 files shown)


dfed:
Seems like a reasonable, non-breaking change. @mikelikespie, look good to you?

mikelikespie:
LGTM. Thanks @jder!

Issue #308, opened 2015-11-16
[Bug] "websocket:didReceiveMessage" don't fired when connection secure


Messme:
Inner method "didConnect" don't fired at all.

fjolnir:
This broke in https://github.com/square/SocketRocket/pull/231
If there are no pinned certificates, then the socket remains in "connecting" state forever.

dfed:
Won't this break the issue #231 was fixing? Looks like we'll still connect before checking the cert.

PR #309, opened 2015-11-17
Fixed connecting limbo

fjolnir:
This fixes an issue where a socket would be stuck in "connecting" state if using SSL without a pinned certificate.

Resolves https://github.com/square/SocketRocket/issues/308

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -1478,9 +1478,7 @@ static const size_t SRFrameHeaderOverhead = 32;
                     [self _failWithError:[NSError errorWithDomain:SRWebSocketErrorDomain code:23556 userInfo:[NSDictionary dictionaryWithObject:[NSString stringWithFormat:@"Invalid server cert"] forKey:NSLocalizedDescriptionKey]]];
                 });
                 return;
-            }
-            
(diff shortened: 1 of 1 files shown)


dfed:
Won't this break the issue #231 was fixing? Looks like we'll still connect before checking the cert.

fjolnir:
`-didConnect` triggers the GET, correct?
As of this PR it should be called as a result of NSStreamEventOpenCompleted only if the url request specifies no pinned certs. If it does, then it should be called after verification (line 1483)

Am I mistaken?

fjolnir:
@dfed I don't want to be pushy here, but currently HEAD is completely broken when using SSL without pinned certs.

NVentimiglia:
Ran into this issue, Thanks @fjolnir

dfed:
Took another pass at this (sorry for the wait folks). This indeed looks good.

Only blocker is @fjolnir needs to sign our [CLA](https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1). Let us know when you've signed, and we'll merge.

fjolnir:
Signed :+1:

dfed:
Thanks! Merging.

PR #311, opened 2015-11-24
Fix OS X framework Carthage support

soffes:
Now it defines a module and has an umbrella header. Before this change, you can't use the OS X version via Carthage.


dfed:
Curious why this PR is failing CI

dfed:
Any updates here? We've now got conflicts too.

dfed:
Closing this out. Feel free to submit a new PR. We'd love to get Carthage working on OS X, but as we don't use SocketRocket on OS X we're relying on the community :).

PR #312, opened 2015-11-26
Setting networkServiceType on stream

MihaMajcen:
Added option to set streamNetworkServiceType on socket. Value is copied from NSURLRequest passed in init.

This implements feature requested in Issue #95 and also adds option to set other types.


MihaMajcen:
I'm sorry, I just found out that this is already implemented in master. I was using latest tag and there is no networkServiceType support.

PR #315, opened 2015-12-04
Fixes an issue where socket would not connect over SSL.

BradleyMorgan:
I might be missing something, but I could not get my app to connect to the socket server using SSL without making this change.  Just thought I would pass it along. Thanks!


fjolnir:
PR for same issue: https://github.com/square/SocketRocket/pull/309

dfed:
Closing this per @fjolnir's comment. We'll take #309 as soon as the CLA is signed.

PR #318, opened 2015-12-15
- Fix OS X build by checking if OS is iOS 9 only on iOS target OSs.

miguelsanchez:
OS X framework build is broken, this patch fixes it by checking if the current OS is iOS 9 only when building for iOS


dfed:
`if` >> `ifdef`, always! Thanks for the fix!

PR #327, opened 2016-01-22
Make SocketRocket usable as a module from within Swift, especially on OS X

davedelong:
Without the umbrella header and the "defines module" build setting, SocketRocket can't be used from within Swift on OS X.


dfed:
This looks reasonable to me! Thanks for taking care of this for us. Merging.

PR #328, opened 2016-01-24
Evalute ssl cert chain with hostname policy for leaf cert.

estobbart:
Without pinning certs, evaluate trust by creating an ssl hostname policy.

### DIFF
--- a/SocketRocket/SRWebSocket.h
+++ b/SocketRocket/SRWebSocket.h
@@ -138,6 +138,7 @@ extern NSString *const SRHTTPResponseErrorKey;
 
 @interface NSMutableURLRequest (SRCertificateAdditions)
 
+// An array of SecCertificateRef's
 @property (nonatomic, retain) NSArray *SR_SSLPinnedCertificates;
 
 @end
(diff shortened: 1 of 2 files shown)


estobbart:
Some helpful docs (for me at least).
https://developer.apple.com/library/mac/qa/qa1360/_index.html
and.. 
https://developer.apple.com/library/ios/technotes/tn2232/_index.html

estobbart:
Any feedback on this? Ideally I'd replace the pinned certificates data compare with a method that evaluated trust by setting the anchor certificate, but I felt that would be a breaking change. If there's any changes you'd like to see here I'm happy to make them.

dfed:
@emonti thoughts here? Someone with an eye for security should take a look at this.

estobbart:
Just of note here. The unspecified result that Apple mentions seems to occur more often with wildcard certs. I've done some testing in this area using SSLHandshake and a native socket. If you don't set a peer name you get the unspecified result on the first trust eval. If you do set a peer name, it comes back as recoverable, you can then create a new policy with the domain and eval it yourself. Either way you end up with unspecified. I included the policy creation step in case for some reason peer name gets set.

estobbart:
Note, this PR has pivoted to take on more of the underlying certificate processing, and probably includes some breaking changes. Consider it a work in progress, and open to additional feedback here.

estobbart:
@nlutsenko Should I close this?

nlutsenko:
@estobbart, yes, I think so.
We have pluggable security policies right now with #429 and help from @michaelkirk.

PR #329, opened 2016-01-25
Update close status codes based on spec.

benshan:
https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent


dfed:
Thanks for updating! Merging.

Issue #333, opened 2016-02-11
Changes to readyState cause too many KVO messages to be sent


danblakemore:
The implementation of the setter for readyState (`- (void)setReadyState:(SRReadyState)aReadyState`) calls `willChangeValueForKey:` and `didChangeValueForKey:` unnecessarily it seems.  

I'm getting duplicate calls to `observeValueForKeyPath:...` and the cause seems to be that the socket is explicitly calling the KVC methods and the system is implicitly duplicating this behavior.  Commenting out the calls to `willBlah` and `didBlah` calls `observeBlah` just once.  [This Stack Overflow answer](http://stackoverflow.com/a/4453594) is what led me to trying to comment those out.

If there's a reason that those calls are explicitly there, I'd be interested to know.  Otherwise, it might be good to remove them because duplicate can break KVO code that assumes there will be one call per actual change (which is most of it, I would assume).

mikelikespie:
@danblakemore seems pretty reasonable. If you submit a PR, we can deff get it in. Bonus points for a unit test that verifies the behavior. Just make sure you sign our [CLA as described here](https://github.com/square/SocketRocket/blob/master/Contributing.md).

nlutsenko:
Looks like this is fully fixed with #335.

PR #335, opened 2016-02-18
Removed duplicated KVO calls for readyState

danblakemore:
Fixed issue #333.  No tests at this point because I can't get the tests to build on my system =D.

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -375,10 +375,8 @@ static __strong NSData *CRLFCRLF;
 
 - (void)setReadyState:(SRReadyState)aReadyState;
 {
-    [self willChangeValueForKey:@"readyState"];
     assert(aReadyState > _readyState);
     _readyState = aReadyState;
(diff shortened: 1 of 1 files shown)


dfed:
Thanks for the PR! Curious what your system is, as unit tests build just fine for me.

danblakemore:
`make test` fails with undefined symbols for architecture i386.  The SocketRocketTests target in Xcode builds, but it didn't look like there was anything in there but the autobahn fuzzing stuff and nothing like a set of tests to add one to.  I thought about at least putting some functionality in the TestChat app that would break and log if the socket sent too many or too few (after my commit) KVO messages, but since `make test` fails, there's not server.

If there is a test added in the future, it should probably test that sequential calls to `observeValueForKeyPath:ofObject:change:context:` have new values considering the `assert(aReadyState > _readyState);` requirement in that setter.

hamchapman:
`make test` fails on master for me to - I started having a go at making the tests run locally for me: https://github.com/square/SocketRocket/pull/330

Issue #325, opened 2016-01-18
ping/pong race condition


nspavlo:
Hello!
There is an issue with messages sent with the following intervals (PING 50ms, DATA 50ms).
Steps to reproduce:
Send ping without payload (0) to SocketRocket client.
Send data (JSON == 188) to SocketRocket client.
After some time SocketRocket client responds to ping with pong containing payload from data frame (JSON == 188).
Repeat till server fails because of received control frame > 126.

Using iOS 9, iPhone 6
SocketRocket (0.4.2) CocoaPods

adematalay:
frame data should be copied to avoid this race condition. When data is received the value at the pointer is changed and pong message has a big body. Thus remote side fails.

galambalazs:
I ran into this exact same issue.

I used 'ws' node library and a ping interval of 30 seconds so this issue happened at seemingly random times.

Then I found out that SocketRocket is sending back pongs with big payloads. The payload in my case seemed to be always the previous text message's payload.

    SR received: 'some text message'
    SR received: ping with empty payload
    SR sent back: pong with payload  'some text message'

galambalazs:
I'll keep testing this, but for the record it seems to be not happening if the PING payload coming from the server has even one character (a space for example).

hashok:
@galambalazs You are right, sending any non-empty payload in the ping towards SocketRocket client makes this issues not to happen. This is the exact workaround I have done on a server side which communicates with the iOS app developed by @nspavlo (we are colleagues).

adematalay:
I created a pull request to fix this issue. If you want you may try my branch.

galambalazs:
@adematalay I saw your pull, not sure I understand why the value at the pointer would *only change* if the ping payload is empty. If it is a race condition and pointer problems the value should change even if the payload has 1 character for example. Right?

Care to explain maybe? :)

adematalay:
When you send ping+data (data can be empty) it sends pong+data back on other thread. data value is always same memory block and while you are sending pong on other thread if any data is received before you fired your pong+data value will be changed. That is why you see the data you sent in the pong. 

If you send empty data in the middle it changes the CPU time so race condition does not happen.

dfed:
This should be fixed on master.

galambalazs:
cool :+1:

nspavlo:
Thanks!

Sent from Outlook Mobile




[...]

PR #338, opened 2016-03-08
frameData is copied before passing to handlers

adematalay:
This is fix of issue #325


dfed:
Thanks for the fix! Merging.

PR #341, opened 2016-03-18
Always return NSData in didReceiveMessage: delegate method

pm-dev:
This is an optimization for a common use case where the server is sending text serialized as JSON. Having socket rocket convert the NSData to an NSString is undesirable because the delegate will be immediately converting back into NSData before passing that to NSJSONSerialization. With this pull request, the client would avoid two data <--> string conversions.


dfed:
You're proposing a breaking change to the API. If you can do this in an non-breaking way, we'll consider (correct me if I'm wrong @mikelikespie). But the current approach is not one we're willing to take.

On another note, I'm not sure why the error handling in the SROpCodeTextFrame case has been removed

pm-dev:
Correct, it's a breaking change, I'll explore an option that is not breaking. Perhaps a call to the delegate to see if it should do the conversion. Delegate could implement optionally.
As for the error handling, if we don't turn the NSData into an NSString, we wont know if the NSData is a valid UTF-8 string.

PR #342, opened 2016-03-18
Message handling optimization

pm-dev:
Here's a non-breaking version of the optimization for a common use case where the server is sending text serialized as JSON. Having socket rocket convert the NSData to an NSString is undesirable because the delegate will be immediately converting back into NSData before passing that to NSJSONSerialization. With this pull request, the client would avoid two data <--> string conversions.

Give the delegate the ability to specify whether SRWebSocket should convert NSData to NSString for TEXT messages. This is an optimization for when the server is sending text serialized as JSON. In this case we would turn the string back into NSData before passing it to NSJSONSerialization, which removes 2 data <--> string conversions.


dfed:
I like this. @mikelikespie can you sanity-check me?

mikelikespie:
:+1:

dfed:
Thanks for the fast turn-around!

pm-dev:
Thanks for picking it up! :+1:

PR #359, opened 2016-05-06
Fix no ability to build tvOS framework for Apple TV Simulator from Xcode.

nlutsenko:
This should be set to `3` for Apple TV Simulator to be targetable.
`1` stands for iPhone/iTouch
`2` stands for iPad


facebook-github-bot-7:

PR #360, opened 2016-05-06
Split internal classes into their own source files.

nlutsenko:
Split and rewritten a good portion of internal consumer classes.
Added todos where relevant.


facebook-github-bot-7:

PR #361, opened 2016-05-06
Update README URLs based on HTTP redirects

ReadmeCritic:
Created with https://github.com/dkhamsing/frankenstein

### GitHub Corrected URLs 
Was | Now 
--- | --- 
https://github.com/facebook/tornado | https://github.com/tornadoweb/tornado 

### HTTPS Corrected URLs 
Was | Now 
--- | --- 
http://tools.ietf.org/html/rfc6455 | https://tools.ietf.org/html/rfc6455


facebook-github-bot-7:
@nlutsenko would you mind taking a look at this pull request? It's been a while since the last commit was reviewed.

PR #363, opened 2016-05-07
Update and unify all product bundle identifiers.

nlutsenko:


facebook-github-bot-0:

PR #364, opened 2016-05-07
Change minimum deployment target to iOS 6.0, OS X 10.8.

nlutsenko:
This is a breaking change if you are stil lusing iOS 5.0 or 10.7.
Changing it gives us an ability to move faster, as well as be in sync with our Podspec and latest Xcode (5.1 is not an option anymore).


facebook-github-bot-0:

hamchapman:
I'm very much in support of this!

PR #362, opened 2016-05-06
Improve Xcode-based Autobahn testing.

nlutsenko:

PR #365, opened 2016-05-07
Add parameterized configuration to autobahn tests.

nlutsenko:
Depends on #362.
Added `autobahn_configuration.json` that contains the expected test type to test identifier prefix.
Allows all of our autobahn tests to pass via Xcode, so we can eventually enable autobahn testing via Xcode.


facebook-github-bot-2:

PR #367, opened 2016-05-10
Make all tests runtime and Xcode discoverable.

nlutsenko:
This makes all of the tests runtime discoverable as well as gives a nice view for all of them in Xcode.
Also, having an implementaiton for `testCaseWithSelector:` gives us an ability to run one or just a few tests only via Xcode, as well as repeatedly run a single test.

Thank you, @modocache, for inspiration on this.


nlutsenko:
Ow, yes, yes, screenshot. The required attribute on anything visually important!
screen shot 2016-05-09 at 8 24 43 pm

PR #368, opened 2016-05-10
Add 'sendData:', 'sendString:' and deprecate 'send:'.

nlutsenko:
Both new methods accept `nil` and are sending it as a text frame (the same behavior as before).
The old method is deprecated now, whee!

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -651,18 +651,35 @@ static __strong NSData *CRLFCRLF;
 
 - (void)send:(id)data;
 {
+    if (!data) {
+        [self sendData:nil]; // Send Data, but it doesn't matter since we are going to send the same text frame with 0 length.
+    } else if ([data isKindOfClass:[NSString class]]) {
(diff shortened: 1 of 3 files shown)


SocketRocket/SRWebSocket.m
```diff
@@ -651,18 +651,33 @@ - (void)_writeData:(NSData *)data;
...
+{
+    NSAssert(self.readyState != SR_CONNECTING, @"Invalid State: Cannot call send: until connection is open");
+    dispatch_async(_workQueue, ^{
+        [self _sendFrameWithOpcode:SROpCodeTextFrame data:[string dataUsingEncoding:NSUTF8StringEncoding]];
+    });
+}
+
+- (void)sendData:(NSData *)data
+{
     NSAssert(self.readyState != SR_CONNECTING, @"Invalid State: Cannot call send: until connection is open");
```
richardjrossiii: We should still copy data... If it's a NSMutableData, doing this async means that unintended changes after a call to `sendData:` has been made. Same with `sendString:`.

PR #370, opened 2016-05-10
Handle very large payload length gracefully on 32-bit

erikprice:


SocketRocket/SRWebSocket.m
```diff
@@ -47,6 +47,7 @@
 #error SocketRocket must be compiled with ARC enabled
 #endif
 
+#import "TargetConditionals.h"
```
nlutsenko: Nit: Use angular brackets for this import, since it's coming from /usr/include
Aka `#import <TargetConditionals.h>`, also might be a good idea to move it next to other framework imports instead of it being below our custom defines.

nlutsenko:
@erikprice, the change is good, but this doesn't quite solve anything 😁 
Enabling warnings - we should merge that regardless!

Frame size issue - not as easy, because even on 32-bit platform if you can read 4GB worth of data from a frame (that also runs on a Web Socket) - that would be fantastic!
What I would suggest doing is:
- Use `uint64_t` instead of `size_t` for the data length
- Handle the error on allocation level instead of on a

nlutsenko:
What would you say if we make this PR about warnings and change of data length type to `uint64_t`?

erikprice:
@nlutsenko thanks for your reply. I'll talk to the original developer who made these changes and will follow up on this issue next week.

erikprice:
These changes are from several years ago, but here is the response I got…

> I think those changes were to make the code compile for 64 bit architectures and the size check was to defend agains buffer overruns.

It looks like if we change `_addConsumerWithDataLength:callback:` to take a `uint64_t` instead of a `size_t` it will ripple a bit through the other classes, as the data length is propagated down to the consumers. Is that what you suggest, @nlutsenko?

nlutsenko:
The data length is not propagated through to the public API, but indeed propagates to consumers in the API. I've landed a couple of fixes for the second portion of this, specifically #372 that should help with failing when we fail to allocate memory for the buffer.
All-in-all, I think it's fair to simply update entire pipeline that relies on `dataLength` to be `uint64_t` instead of `size_t`.

nlutsenko:
Closing, as this should be handled gracefully now per #372.
Let me know if you still can encounter this problem.

Issue #114, opened 2013-09-16
Feature request. iOS devices. Work through http proxy


shyswork:
It would be wonderful if you could implement work through http proxy. Right now if iOS device configured to work through http proxy and without DNS, SocketRocket fails with error 64(host is down) or 60(Operation timed out). I am not sure if this is that common for iOS devices to have such configuration but anyway.
![sr_example](https://f.cloud.github.com/assets/3229954/1150233/6f90c9b4-1ee7-11e3-8649-d06a15003caa.png)

mikelikespie:
Doing some housekeeping. Please reopen if still relevant

dfed:
Reopened due to feedback

madlymad:
> @msgilligan: You could try using WSS/HTTPS.

In my case the not the WS/HTTP nor the WSS/HTTPS work when I am behind the proxy.

rahulsh1:
+1. Can't get the app working within the corporate network using the proxy.

diwakargoel:
+1

danieljimenez:
+1

MarekR22:
When I'm testing this feature with latest `master` it works fine:
- no proxy ws and wss
- proxy and ws

and it doesn't work in case of:
- unsecured proxy and wss connection, I get certificate error. (everything works when bypassing proxy).

PR #395, opened 2016-05-30
Add Proxy support

oldshuren:
This pull request added support to use SocketRocket behind http proxy.

The proxy setting is read from the system configuration. PAC is also supported.

If there is proxy in the system configuration. Instead open network stream to the web socket server,   the network stream to the proxy server is opened. Then we send HTTP CONNECT with the web socket server address to the proxy server. After receiving 200 OK status, it operates like normal network stream.

### DIFF
--- a/SocketRocket/ProxyConnect.h
+++ b/SocketRocket/ProxyConnect.h
@@ -1,6 +1,0 @@
-#import <Foundation/Foundation.h>
-
-@interface ProxyConnect : NSObject
--(instancetype)initWithURL:(NSURL *)url;
--(void) openNetworkStreamWithCompletion:(void (^)(NSError *error, NSInputStream *readStream, NSOutputStream *writeStream ))completion;
-@end								  
--- a/SocketRocket/ProxyConnect.h
+++ b/SocketRocket/ProxyConnect.h
@@ -1,0 +1,6 @@
+#import <Foundation/Foundation.h>
+
+@interface ProxyConnect : NSObject
+-(instancetype)initWithURL:(NSURL *)url;
+-(void) openNetworkStreamWithCompletion:(void (^)(NSError *error, NSInputStream *readStream, NSOutputStream *writeStream ))completion;
+@end								  
--- a/SocketRocket/ProxyConnect.m
+++ b/SocketRocket/ProxyConnect.m
@@ -1,0 +1,468 @@
+#import "ProxyConnect.h"
+#import "SRError.h"
+#import "NSRunLoop+SRWebSocket.h"
+
+static inline void ProxyFastLog(NSString *format, ...);
+
+typedef void (^connectDoneBlock_t)(NSError * error, NSInputStream *readStream, NSOutputStream *writeStream);
+
+@interface ProxyConnect() <NSStreamDelegate>
+
+@property (nonatomic, strong) NSURL *url;
+@property (nonatomic, strong) NSInputStream * inputStream;
+@property (nonatomic, strong) NSOutputStream * outputStream;
+
+@end
+@implementation ProxyConnect {
+    
+    connectDoneBlock_t  connectDoneHandler;
+    
+    NSString *_httpProxyHost;
+    uint32_t _httpProxyPort;
+    
+    CFHTTPMessageRef _receivedHTTPHeaders;
+
+    NSString *_socksProxyHost;
+    uint32_t _socksProxyPort;
+    NSString *_socksProxyUsername;
+    NSString *_socksProxyPassword;
+    
+    BOOL _secure;
+    
+    NSMutableArray<NSData *> *_inputQueue;
+    NSOperationQueue * _writeQueue;
+    
+}
+
+-(instancetype)initWithURL:(NSURL *)url
+{
+    if ([super init]) {
+        self.url = url;
+        
+        NSString *scheme = url.scheme.lowercaseString;
+        
+        if ([scheme isEqualToString:@"wss"] || [scheme isEqualToString:@"https"]) {
+            _secure = YES;
+        }
+        
+        _writeQueue =  [[NSOperationQueue alloc] init];
+        _inputQueue = [NSMutableArray arrayWithCapacity:2];
+    }
+    return self;
+}
+
+-(void) openNetworkStreamWithCompletion:(void (^)(NSError *error, NSInputStream *readStream, NSOutputStream *writeStream ))completion
+{
+    connectDoneHandler = completion;
+    [self _configureProxy];
+}
+
+-(void) _didConnect
+{
+    ProxyFastLog(@"_didConnect, return streams");
+    if (_secure) {
+        if (_httpProxyHost) {
+            // Must set the real peer name before turning on SSL
+            ProxyFastLog(@"proxy set peer name to real host %@", self.url.host);
+            [self.outputStream setProperty:self.url.host forKey:@"_kCFStreamPropertySocketPeerName"];
+        }
+    }
+    if (_receivedHTTPHeaders) {
+        CFRelease(_receivedHTTPHeaders);
+        _receivedHTTPHeaders = NULL;
+    }
+    NSInputStream *inputStream = self.inputStream;
+    NSOutputStream *outputStream = self.outputStream;
+    self.inputStream = nil;
+    self.outputStream = nil;
+    [inputStream removeFromRunLoop:[NSRunLoop SR_networkRunLoop]
+                           forMode:NSDefaultRunLoopMode];
+    inputStream.delegate = nil;
+    outputStream.delegate = nil;
+    connectDoneHandler(nil, inputStream, outputStream);
+}
+
+-(void) _failWithError:(NSError *)error
+{
+    ProxyFastLog(@"_failWithError, return error");
+    if (!error) {
+        error = SRHTTPErrorWithCodeDescription(500, 2132,@"Proxy Error");
+    }
+    
+    if (_receivedHTTPHeaders) {
+        CFRelease(_receivedHTTPHeaders);
+        _receivedHTTPHeaders = NULL;
+    }
+    [self.inputStream removeFromRunLoop:[NSRunLoop SR_networkRunLoop]
+                                forMode:NSDefaultRunLoopMode];
+    [self.inputStream close];
+    [self.outputStream close];
+    self.inputStream = nil;
+    self.outputStream = nil;
+    connectDoneHandler(error, nil, nil);
+}
+
+// get proxy setting from device setting
+-(void) _configureProxy
+{
+    ProxyFastLog(@"configureProxy");
+    NSDictionary *proxySettings = CFBridgingRelease(CFNetworkCopySystemProxySettings());
+    // CFNetworkCopyProxiesForURL doesn't understand ws:// or wss://
+    NSURL *httpURL;
+    if (_secure)
+        httpURL = [NSURL URLWithString:[NSString stringWithFormat:@"https://%@", _url.host]];
+    else
+        httpURL = [NSURL URLWithString:[NSString stringWithFormat:@"http://%@", _url.host]];
+    
+    NSArray *proxies = CFBridgingRelease(CFNetworkCopyProxiesForURL((__bridge CFURLRef)httpURL, (__bridge CFDictionaryRef)proxySettings));
+    if (proxies.count == 0) {
+        ProxyFastLog(@"configureProxy no proxies");
+        [self _openConnection];
+        return;                 // no proxy
+    }
+    NSDictionary *settings = [proxies objectAtIndex:0];
+    NSString *proxyType = settings[(NSString *)kCFProxyTypeKey];
+    if ([proxyType isEqualToString:(NSString *)kCFProxyTypeAutoConfigurationURL]) {
+        NSURL *pacURL = settings[(NSString *)kCFProxyAutoConfigurationURLKey];
+        if (pacURL) {
+            [self _fetchPAC:pacURL];
+            return;
+        }
+    }
+    if ([proxyType isEqualToString:(NSString *)kCFProxyTypeAutoConfigurationJavaScript]) {
+        NSString *script = settings[(NSString *)kCFProxyAutoConfigurationJavaScriptKey];
+        if (script) {
+            [self _runPACScript:script];
+            return;
+        }
+    }
+    [self _readProxySettingWithType:proxyType settings:settings];
+
+    [self _openConnection];
+}
+
+- (void) _readProxySettingWithType:(NSString *)proxyType settings:(NSDictionary *)settings
+{
+    if ([proxyType isEqualToString:(NSString *)kCFProxyTypeHTTP] || [proxyType isEqualToString:(NSString *)kCFProxyTypeHTTPS]) {
+        _httpProxyHost = settings[(NSString *)kCFProxyHostNameKey];
+        NSNumber *portValue = settings[(NSString *)kCFProxyPortNumberKey];
+        if (portValue)
+            _httpProxyPort = [portValue intValue];
+    }
+    if ([proxyType isEqualToString:(NSString *)kCFProxyTypeSOCKS]) {
+        _socksProxyHost = settings[(NSString *)kCFProxyHostNameKey];
+        NSNumber *portValue = settings[(NSString *)kCFProxyPortNumberKey];
+        if (portValue)
+            _socksProxyPort = [portValue intValue];
+        _socksProxyUsername = settings[(NSString *)kCFProxyUsernameKey];
+        _socksProxyPassword = settings[(NSString *)kCFProxyPasswordKey];
+    }
+    if (_httpProxyHost) {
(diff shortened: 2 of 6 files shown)


facebook-github-bot-0:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

oldshuren:
Signed GLA

facebook-github-bot-6:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

SocketRocket/SRWebSocket.m
```diff
@@ -145,6 +145,122 @@ @implementation SRWebSocket {
...
+
+    //Proxy support
+    NSString *_httpProxyHost;
+    uint32_t _httpProxyPort;
+    BOOL _connectingToProxy;
+
+}
+
+// get proxy setting from device setting
+-(void) configureProxy
```
nlutsenko: Nit: Please match the overall style - aka `- (void)_configureProxy`

SocketRocket/SRWebSocket.m
```diff
@@ -145,6 +145,122 @@ @implementation SRWebSocket {
...
+        NSString *script = [NSString stringWithContentsOfURL:PACurl
+                                                usedEncoding:NULL
+                                                       error:&nsError];
+
+        if (nsError) {
+            [self openConnection];
+            return;
+        }
+        [self runPACScript:script];
+		return;
```
nlutsenko: Nit: Looks like this is missing indentation - if you could run `auto-indent` in Xcode on all of the new code - that would help tremendously.

SocketRocket/SRWebSocket.m
```diff
@@ -145,6 +145,122 @@ @implementation SRWebSocket {
...
+        // Don't know how to read data from this URL, we'll have to give up
+        // We'll simply assume no proxies, and start the request as normal
+        [self openConnection];
+		return;
+	}
+    __weak typeof(self) weakSelf = self;
+    NSURLRequest *request = [NSURLRequest requestWithURL:PACurl
+                                             cachePolicy:NSURLRequestReturnCacheDataElseLoad
+                                         timeoutInterval:120.0];
+    [NSURLConnection sendAsynchronousRequest:request queue:[NSOperationQueue mainQueue]
```
nlutsenko: This one is important:
NSURLConnection is deprecated as of this point, and since we support iOS 6.0+ we can safely use NSURLSession for this request.

SocketRocket/SRWebSocket.m
```diff
@@ -145,6 +145,122 @@ @implementation SRWebSocket {
...
+	NSString *scheme = [PACurl.scheme lowercaseString];
+    if (![scheme isEqualToString:@"http"] && ![scheme isEqualToString:@"https"]) {
+        // Don't know how to read data from this URL, we'll have to give up
+        // We'll simply assume no proxies, and start the request as normal
+        [self openConnection];
+		return;
+	}
+    __weak typeof(self) weakSelf = self;
+    NSURLRequest *request = [NSURLRequest requestWithURL:PACurl
+                                             cachePolicy:NSURLRequestReturnCacheDataElseLoad
```
nlutsenko: Is there a specific reason on why you would want to use `CacheDataElseLoad`? If yes - could you please add a comment, so the next person looking at it can figure this out.

SocketRocket/SRWebSocket.m
```diff
@@ -281,6 +395,27 @@ - (BOOL)_checkHandshake:(CFHTTPMessageRef)httpMessage;
     return [acceptHeader isEqualToString:expectedAccept];
 }
 
+- (void)_ProxyHTTPHeadersDidFinish;
```
nlutsenko: Nit: Please match overall style aka `- (void)_proxyHTTPHeadersDidFinish`.

SocketRocket/SRWebSocket.m
```diff
@@ -452,10 +629,10 @@ - (void)_updateSecureStreamOptions;
         [_outputStream setProperty:sslOptions forKey:(__bridge NSString *)kCFStreamPropertySSLSettings];
     }
     
-    _inputStream.delegate = self;
-    _outputStream.delegate = self;
+    //_inputStream.delegate = self;
```
nlutsenko: This looks like an unintentional change. Any chance you can double check that we still assign the `delegate `for the streams?

nlutsenko:
@oldshuren, than you so much for figuring out the CFNetwork approach to auto-proxy configuration.
The PR looks like a really great start. I put some nitpicky notes on it, please note that all of these are applicable to the entire code change.

Also, if we could make the proxy configuration a self contained set of functions, or even a self contained class - that would be much better for code maintainability and reuse.

madlymad:
I can verify that the patch works for the HTTPS connection through proxy that is set with the pac file URL in the device's WiFi settings. The corporate proxy does not support socks connections.

Tested on iPhone 4S (iOS 9.3.1 build 13E283).

SocketRocket/SRWebSocket.m
oldshuren: Ok,

I'll do it.

Dong

SocketRocket/SRWebSocket.m
oldshuren: Ok, I'll change it.

Dong

SocketRocket/SRWebSocket.m
oldshuren: Yes, I think NSURLRequestReloadRevalidatingCacheData is better. I think 
the timeout value is also too long.

Dong

oldshuren:
I think we can make a class to handle the proxy configuration and 
connecting to the http server. The completion block will be invoke with 
either two input/output stream for success, or NSError for error.

BTW, if I make the change, do I need to submit a new pull request?

Thanks,

Dong

nlutsenko:
> I think we can make a class to handle the proxy configuration and
connecting to the http server. The completion block will be invoke with
either two input/output stream for success, or NSError for error.

Sounds great. Looking forward to this!

> BTW, if I make the change, do I need to submit a new pull request?

You can just force push the existing remote branch with `git push -f` which will update the pull request here automatically.

oldshuren:
Or create a class called NetSocket, which will handle proxy, open 
network connection, send and receiving raw data?

Dong

nlutsenko:
I would say - we are good just by using a single class for proxy setup here, I wouldn't advise a big change, like NetSocket implementation just yet, since it's slightly out of scope of the intention here (http proxy handling).

gaborvass:
Hi,

Would be possible to add SOCKS proxy support as well? 

Thanks,
Gabor

oldshuren:
I guess it is possible, but you have to use PAC, because you can not 
input SOCK directly in iOS.

I'll take a look after I finish the HTTP proxy.

Cheers,

Dong

gaborvass:
you can add it on ios like this:

`CFMutableDictionaryRef socksConfig = CFDictionaryCreateMutableCopy(NULL, 0, proxyDict);`
`CFStringRef socksProxyHost = CFDictionaryGetValue(proxyDict, @"SOCKSProxy");`
`CFNumberRef socksProxyPort = CFDictionaryGetValue(proxyDict, @"SOCKSPort");`
`if (socksProxyHost != NULL && socksProxyPort != NULL)`
`{`
`            CFDictionarySetValue(socksConfig, kCFStreamPropertySOCKSProxyHost, socksProxyHost);`
`            CFDictionarySetValue(socksConfig, kCFStreamPropertySOCKSProxyPort, socksProxyPort);`
`            CFDictionarySetValue(socksConfig, kCFStreamPropertySOCKSVersion, kCFStreamSocketSOCKSVersion4);`
`            CFReadStreamSetProperty(readStream, kCFStreamPropertySOCKSProxy, proxyDict);`
`            CFWriteStreamSetProperty(writeStream, kCFStreamPropertySOCKSProxy, proxyDict);`
`            CFRelease(socksProxyHost);`
`            CFRelease(socksProxyPort);`
`        }`
`        CFRelease(socksConfig);`

cheers,
Gabor

oldshuren:
I pushed new update. Proxy handling was moved to ProxyConnect.{h|m}

Socks support is added. It turns out Socks support is easier that HTTP, Apple's CFNetwork already have Socks support, you just need to configuration from system settings.

oldshuren:
@gaborvass I did it just like your suggestion, but on the NSStream. And the system configure only provide host, port, username and password. So I just ignore the version attribute.

I tested it with the ssh Socks tunnel, it works perfectly.

gaborvass:
@oldshuren thanks a lot!

SocketRocket/ProxyConnect.h
```diff
@@ -0,0 +1,6 @@
+#import <Foundation/Foundation.h>
+
+@interface ProxyConnect : NSObject
```
nlutsenko: Let's rename this class, so we are consistent and can avoid clashing with anyone who has `ProxyConnect` in their app. Suggested name: `SRProxyConnect` :)

SocketRocket/ProxyConnect.h
```diff
@@ -0,0 +1,6 @@
...
+-(instancetype)initWithURL:(NSURL *)url;
```
nlutsenko: Surround the declaration of `init` and `openNetworkStream` with 2 empty lines, per common style.

SocketRocket/ProxyConnect.h
```diff
@@ -0,0 +1,6 @@
+#import <Foundation/Foundation.h>
```
nlutsenko: This is missing the copyright declaration, which is required for all our source files.
Please copy it from `SRError.h`

SocketRocket/ProxyConnect.m
```diff
@@ -0,0 +1,468 @@
+#import "ProxyConnect.h"
+#import "SRError.h"
+#import "NSRunLoop+SRWebSocket.h"
+
+static inline void ProxyFastLog(NSString *format, ...);
+
+typedef void (^connectDoneBlock_t)(NSError * error, NSInputStream *readStream, NSOutputStream *writeStream);
```
nlutsenko: Nit: Suggested block type name: `SRProxyConnectionCompletion`.

nlutsenko: Nit: Also, please attach the `*` to the parameter name. ;)

SocketRocket/ProxyConnect.m
```diff
@@ -0,0 +1,468 @@
...
+
+@interface ProxyConnect() <NSStreamDelegate>
+
+@property (nonatomic, strong) NSURL *url;
+@property (nonatomic, strong) NSInputStream * inputStream;
+@property (nonatomic, strong) NSOutputStream * outputStream;
+
+@end
+@implementation ProxyConnect {
```
nlutsenko: Nit: Please add an empty line before `@implementation`, per common style.

SocketRocket/ProxyConnect.m
```diff
@@ -0,0 +1,468 @@
...
+    BOOL _secure;
+    
+    NSMutableArray<NSData *> *_inputQueue;
+    NSOperationQueue * _writeQueue;
+    
+}
+
+-(instancetype)initWithURL:(NSURL *)url
+{
+    if ([super init]) {
```
nlutsenko: This would be `self = [super init]` or the common one we use:
```
self = [super init];
if (!self) return self;
```
which will also avoid creating an extra indentation level.

SocketRocket/ProxyConnect.m
```diff
@@ -0,0 +1,468 @@
...
+        self.url = url;
```
nlutsenko: Nit: Please avoid using property setters in constructor aka replace with `_url = url;`.

SocketRocket/ProxyConnect.m
```diff
@@ -0,0 +1,468 @@
+#import "ProxyConnect.h"
+#import "SRError.h"
+#import "NSRunLoop+SRWebSocket.h"
+
+static inline void ProxyFastLog(NSString *format, ...);
```
nlutsenko: Nit: Rename to `SRProxyFastLog()

SocketRocket/ProxyConnect.m
```diff
@@ -0,0 +1,468 @@
...
+    // Send HTTP CONNECT Request
+    NSString *connectRequestStr = [NSString stringWithFormat:@"CONNECT %@:%u HTTP/1.1\r\nHost: %@\r\nConnection: keep-alive\r\nProxy-Connection: keep-alive\r\n\r\n", _url.host, port, _url.host];
+    
+    NSData *message =  [connectRequestStr dataUsingEncoding:NSUTF8StringEncoding];
+    ProxyFastLog(@"Proxy sending %@", connectRequestStr);
+    
+    [self _writeData:message];
+}
+
+#define BUFFER_MAX  4096
```
nlutsenko: Nit: Replace with constant aka `static size_t const SRProxyConnectBufferMaxSize = 4096;`

SocketRocket/ProxyConnect.m
```diff
@@ -0,0 +1,468 @@
...
+@interface ProxyConnect() <NSStreamDelegate>
+
+@property (nonatomic, strong) NSURL *url;
+@property (nonatomic, strong) NSInputStream * inputStream;
+@property (nonatomic, strong) NSOutputStream * outputStream;
+
+@end
+@implementation ProxyConnect {
+    
+    connectDoneBlock_t  connectDoneHandler;
```
nlutsenko: Nit: Prefix with underscore per common style.

nlutsenko:
@oldshuren, this looks good overall, few nits here and there (there is more but I skipped them for now).

Overall notes:
- Runloop schedule/unschedule is hacky, but will work just as fine.
- Love the fact that we decoupled this code from the `SRWebSocket` itself, it's way more manageable this way!
- We have a thread safety problem here, because the completion block is not guaranteed to be called on the same thread as it was scheduled, you might want to wrap the completion either in the work queue scheduling.
- Since the root folder of `SocketRocket` is intended for public classes only - merging this right now will make it public, which is something we would want to avoid. Can you please move the `ProxyConnect` into `SocketRocket/Internal`?

As per more nits, styling, structuring: 
Would you want to go through few iterations before I merge this? Or would it be more helpful if we merge this after fixing few things and I'll take a pass on structuring/linting/adding thread safety and unifying the style here?

The minimum required changes to merge this:
- Adding copyrights
- Renaming `ProxyConnect` to `SRProxyConnect`
- Moving `SRProxyConnect` into `SocketRocket/Internal`

Let me know how I can support you here.

oldshuren:
Initially I had the stream running in the default Runloop,  the normal 
connection worked fine, but the connection to the proxy server didn't 
work. So I'm just using SRWebSocket own Runloop.

You don't need to unschedule in the proxy handling code, but it requires 
more change in the main SRWebSocket.m :(

I think I'll do the minimum necessary changes and let you change the 
other part. It is much easier this way.

Thanks,

Dong

oldshuren:
On 6/3/16 2:35 PM, Nikita Lutsenko wrote:
>
>   * We have a thread safety problem here, because the completion block
>     is not guaranteed to be called on the same thread as it was
>     scheduled, you might want to wrap the completion either in the
>     work queue scheduling.
>
I tried to call the completion handler in the invoking thread, that is 
main thread. There seems some race conditions around, it sometimes 
worked, sometime didn't.

The proxy code doesn't have access to the SRWebSocket's worker queue. So 
I cannot wrap it in proxy code. But in the main SRWebSocket, didConnect 
is wrapped in the work queue, as

                 dispatch_async(_workQueue, ^{
                         [weakSelf didConnect];
                     });

So I think it should be fine.

Dong

oldshuren:
One question, does the copyrights have to say Facebook?

Thanks,

Dong

nlutsenko:
@oldshuren, yes, that's an explicit requirement.

oldshuren:
Ok

Dong

nlutsenko:
Perfect, than you for implementing this.
Merging now, and am going to send another pull request that lints the change as well as cleans up some leftovers.

SocketRocket/SRWebSocket.m
```diff
@@ -1570,7 +1567,7 @@ - (NSOperationQueue *_Nullable)delegateOperationQueue
 
 @end
 
-//#define SR_ENABLE_LOG
+#define SR_ENABLE_LOG
```
cgarbin: Hi guys. Great work and apologies in advance for coming late to the party. 

Did you mean to enable logging by default here for all builds?

nlutsenko: Should be disabled on the current master branch.
Let me know if it's not. Shouldn't be enabled by default.

cgarbin: master branch is ok

(should have checked there first - sorry for the wasted time)

happy-tanuki:
This is great job!
Why not the tickets #114 and #395 is closed?

PR #406, opened 2016-06-09
Fix pinned certificates and extract SSL-related things into SRSecurityOptions.

### LINKED ISSUE #402: It  cannot connect with SSL  certification since last update.
It  cannot connect with SSL  certification since last update.

Last week, We checkout the source code and test it using TestChat project to connect our server through wss://.  It's work perfectly!

But, We checkout a new version today. It not work. We merge the source code found it has been changed.

The error code is:

Error Domain=SRWebSocketErrorDomain Code=2133 "Invalid Sec-WebSocket-Accept response." UserInfo={NSLocalizedDescription=Invalid Sec-WebSocket-Accept response.

nlutsenko:
The fix is trivial - we were calling into `didConnect` twice, but if SSL is enabled - we should only call into it after the validation finishes.
The refactoring here should help with reusing code as well as encapsulating it together.

Validates that the fix actually works and security is still there by running against `wss://echo.websocket.org` and `ws://echo.websocket.org` with these options:
- No pinned cert (SSL: on)
- No pinned cert (SSL: off)
- Pinned single cert from the chain of `*.websocket.org`
- Pinned multi cert from the chain of `*.websocket.org`
- Pinned wrong cert (actually just used `Apple WWDR` which is a valid one but is not in the chain)
- Pinned valid and wrong cert (one from `*.websocket.org` and `Apple WWDR`)

Everything looks good and we properly bail-out as soon as we detect that something doesn't validate.

Fixes #402.


richardjrossiii:
Logic makes sense to me.

PR #407, opened 2016-06-09
Enable Autobahn (iOS only) and Carthage tests on Travis-CI.

nlutsenko:
This PR enables our beloved autobahn tests to be run automatically on every single PR in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,29 +2,42 @@ branches:
   only:
     - master
 language: objective-c
+os: osx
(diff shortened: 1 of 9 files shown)


nlutsenko:
Victory!
We have autobahn running in a virtualized environment on Travis and locally!
Lesson learned - never try to install or build anything related to Python (or in general - anything) from inside `xcodebuild` with the environment variables passed from something being already built.

PR #415, opened 2016-06-16
Split creation of CFHTTPMessage and random data into separate source files.

nlutsenko:
Split for the sake of readability and making SRWebSocket class smaller.

### DIFF
--- a/SocketRocket/Internal/Utilities/SRRandom.m
+++ b/SocketRocket/Internal/Utilities/SRRandom.m
@@ -1,0 +1,26 @@
+//
+// Copyright (c) 2016-present, Facebook, Inc.
+// All rights reserved.
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree. An additional grant
+// of patent rights can be found in the PATENTS file in the same directory.
+//
+
+#import "SRRandom.h"
+
+#import <Security/SecRandom.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+NSData *SRRandomData(NSUInteger length)
+{
+    NSMutableData *data = [NSMutableData dataWithLength:length];
+    int result = SecRandomCopyBytes(kSecRandomDefault, data.length, data.mutableBytes);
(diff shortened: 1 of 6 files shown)


SocketRocket/Internal/Utilities/SRRandom.m
```diff
@@ -0,0 +1,26 @@
...
+#import <Security/SecRandom.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+NSData *SRRandomData(NSUInteger length)
+{
+    NSMutableData *data = [NSMutableData dataWithLength:length];
+    int result = SecRandomCopyBytes(kSecRandomDefault, data.length, data.mutableBytes);
+    if (result != 0) {
+        //TODO: (nlutsenko) Check if there was an error.
```
richardjrossiii: This is one of those cases where I would actually think throwing an exception is the right thing to do. If this fails - it means that our entire security stack is likely to be busted.

nlutsenko: Agreed. Will add an exception here.

SocketRocket/SRWebSocket.m
```diff
@@ -345,62 +394,21 @@ - (void)_readHTTPHeader;
...
-        CFHTTPMessageSetHeaderFieldValue(request, (__bridge CFStringRef)key, (__bridge CFStringRef)obj);
-    }];
+    CFHTTPMessageRef message = SRHTTPConnectMessageCreate(_urlRequest,
+                                                          _secKey,
+                                                          SRWebSocketProtocolVersion,
+                                                          self.requestCookies,
+                                                          _requestedProtocols);
 
-    NSData *message = CFBridgingRelease(CFHTTPMessageCopySerializedMessage(request));
+    NSData *messageData = CFBridgingRelease(CFHTTPMessageCopySerializedMessage(message));
```
richardjrossiii: Usually I prefer `__bridge_transfer` as I find it more explicit, but if we're using `CFBridgingRelease` everywhere its fine.

PR #420, opened 2016-06-21
Make SRWebSocket.readyState KV Observable and thread-safe.

### LINKED ISSUE #314: EXC_BAD_ACCESS in closeWithCode
I am attempting to connect to a web socket using this library from within a Unity3d application. Shortly after calling Open I get a crash with a EXC_BAD_ACCESS (code=1, address=0x30) on thread 1. I am not sure how to debug this. Here is the code that Xcode shows me.

`````
- (void)closeWithCode:(NSInteger)code reason:(NSString *)reason;
{
    assert(code);
    dispatch_async(_workQueue, ^{
        if (self.readyState == SR_CLOSING || self.readyState == SR_CLOSED) {
            return;
        }
        
        BOOL wasConnecting = self.readyState == SR_CONNECTING;

`````

I am crashing on the dispatch_async line.

my code is below

`````
    SRWebSocket *ws =[[SRWebSocket alloc] initWithURLRequest:[NSURLRequest requestWithURL:url]];
    ws.delegate = self;
        
    [ws open];
`````

Am I doing it wrong ?

### PR DESCRIPTION
**Pull request by nlutsenko**
Looks like this property is mutated from a `workQueue`, but can be read from any other thread - so make it fully thread-safe as well as implement custom KVO for it, so we synchronize kvo notifications as well.
Fixes #314.

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -25,6 +25,8 @@
 #import <CoreServices/CoreServices.h>
 #endif
 
+#import <libkern/OSAtomic.h>
+
 #import "SRDelegateController.h"
 #import "SRIOConsumer.h"
 #import "SRIOConsumerPool.h"
@@ -38,6 +40,7 @@
 #import "SRHTTPConnectMessage.h"
 #import "SRRandom.h"
 #import "SRLog.h"
+#import "SRMutex.h"
 
 #if !__has_feature(objc_arc)
 #error SocketRocket must be compiled with ARC enabled
@@ -86,7 +89,7 @@ NSString *const SRHTTPResponseErrorKey = @"HTTPResponseStatusCode";
 
 @interface SRWebSocket ()  <NSStreamDelegate>
 
-@property (nonatomic, assign, readwrite) SRReadyState readyState;
+@property (atomic, assign, readwrite) SRReadyState readyState;
 
 // Specifies whether SSL trust chain should NOT be evaluated.
 // By default this flag is set to NO, meaning only secure SSL connections are allowed.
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: SocketRocket/SRWebSocket.m [hunk 1]
```diff
@@ -151,6 +153,8 @@ @implementation SRWebSocket {
     SRProxyConnect *_proxyConnect;
 }
 
+@synthesize readyState = _readyState;
```
erikprice: Curious, what does this `@synthesize` declaration do for this PR?

nlutsenko: Creates an ivar, since both getter+setter are overridden and I would rather not have an explicit ivar added, but attach everything to properties.

nlutsenko:
@erikprice, FYI just updated once again, turns out I forgot to destroy the mutex attr.
Should be good to go for testing.

erikprice:
Thanks, will update to latest.

erikprice:
I have been running our test suites against aabe895 and haven't yet experienced the crash. Yesterday I experienced the crash ~20% of the time, so this is very encouraging. Thanks for your effort on this, @nlutsenko . We're looking forward to the next release that includes this fix.

richardjrossiii:
As, discussed offline, add some thread safety attributes to `SRMutex`, add a second spin lock for iVar access (separate from the KVO mutex).

Passing back to you so I don't review this again :D

SocketRocket/SRWebSocket.m
debugly: can you remove atomic ? when i use infer scan my project, got them:


> Direct access to ivar `_readyState` of an atomic property at line 300, column 13. Accessing an ivar of an atomic property makes the property nonatomic.

PR #422, opened 2016-06-23
Fix CI tests with Carthage 0.17.

nlutsenko:
Apparently 0.17 now requires for Cartfile and Cartfile.resolved to exist, even though they are empty.
Temporary solution until we get more context on Carthage/Carthage#1366.


nlutsenko:
No longer required, since it was fixed with Carthage 0.17.1.

Issue #424, opened 2016-06-24
Delegate methods to avoid copying NSData messages


erikprice:
I have a PR that I would like to submit, which adds `SRWebSocketDelegate` methods to allow the delegate to choose whether to copy the NSData messages received over the web socket:

```
/**
 Called to determine whether to copy outbound data before sending.

 Default (if not implemented) is to copy.

 @param webSocket An instance of `SRWebSocket` that is about to send a message.
 @param data      Data to send in a form of `NSData`.
*/
- (BOOL)webSocket:(SRWebSocket *)webSocket shouldCopyDataToSend:(id)data;

/**
 Called to determine whether to copy inbound data before sending to `-webSocket:didReceiveMessage:`.

 Default (if not implemented) is to copy.

 @param webSocket An instance of `SRWebSocket` that received a message.
 @param data      Received data in a form of `NSData`.
*/
- (BOOL)webSocket:(SRWebSocket *)webSocket shouldCopyReceivedData:(NSData *)data;
```

We have been using a customized version of SocketRocket based on `0.5.0` that uses these methods to disable the copying of the web socket messages, which has brought us noticeable performance gains. I wanted to submit our changes back as a PR, but the `SRDelegateController` API has been added in more recent versions of the library. This calls delegate methods asynchronously, so there doesn't seem to be a preferred way to have delegate methods that return values. So I have two questions:

1. Would the SocketRocket library welcome a PR to bring this functionality? We would prefer not to continue to maintain a custom-patched version of the library in order to have it. Notice that it would be opt-in, and the default behavior would be to remain the same (copy NSData objects).
2. If the answer to 1 is yes, do you have any recommendations on how the delegate object could communicate the policy to the SRWebSocket object?

nlutsenko:
Both of these are a very good addition, but I think we can do better than this, specifically for `shouldCopyDataToSend:`:
Add a method `sendDataNoCopy:error:` in addition to existing one, which will directly relate to how you want the data to be sent.

I am curious on the internal impact of the receiving end as well, since we try our best to never copy anything unless we have to, meaning that if you have an optimization for these - we should merge it without even asking whether someone wants to have a copy of received data or not.

erikprice:
> I am curious on the internal impact of the receiving end as well, since we try our best to never copy anything unless we have to, meaning that if you have an optimization for these - we should merge it without even asking whether someone wants to have a copy of received data or not.

[...]

PR #426, opened 2016-06-27
Make copying data optional

erikprice:
The context for this PR is in #424 .

### DIFF
--- a/SocketRocket/SRWebSocket.h
+++ b/SocketRocket/SRWebSocket.h
@@ -330,7 +330,12 @@ NS_DESIGNATED_INITIALIZER;
  */
 - (void)webSocket:(SRWebSocket *)webSocket didReceiveMessageWithData:(NSData *)data;
 
-// Whether to copy inbound data before sending to -webSocket:didReceiveMessage:. Default (if not implemented) is to copy.
+/**
+ Called when a frame is received from a web socket, to decide whether a defensive copy should be made.
+
+ @param webSocket An instance of `SRWebSocket` that received a message.
+ @param data      Received data in a form of `NSData`.
+ */
 - (BOOL)webSocket:(SRWebSocket *)webSocket shouldCopyReceivedData:(NSData *)data;
 
 #pragma mark Status & Connection
--- a/SocketRocket/SRWebSocket.h
+++ b/SocketRocket/SRWebSocket.h
@@ -266,6 +266,18 @@ NS_DESIGNATED_INITIALIZER;
 - (BOOL)sendData:(nullable NSData *)data error:(NSError **)error NS_SWIFT_NAME(send(data:));
 
 /**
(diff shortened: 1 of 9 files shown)


SocketRocket/SRWebSocket.m
```diff
@@ -997,7 +1014,9 @@ - (void)_readFrameContinue;
 - (void)_readFrameNew;
 {
     dispatch_async(_workQueue, ^{
-        [_currentFrameData setLength:0];
+        // Don't reset the length, since Apple doesn't guarantee that this will free the memory (and in tests on
+        // some platforms, it doesn't seem to, effectively causing a leak the size of the biggest frame so far).
+        _currentFrameData = [[NSMutableData alloc] init];
```
nlutsenko: This by itself is worth a ton!
Indeed this is the only valid way to reset the data...

SocketRocket/SRWebSocket.h
```diff
@@ -264,6 +264,18 @@ NS_DESIGNATED_INITIALIZER;
...
+ Send binary data to the server, without making a defensive copy of it first.
+
+ @param data  Data to send.
+ @param error On input, a pointer to variable for an `NSError` object.
+ If an error occurs, this pointer is set to an `NSError` object containing information about the error.
+ You may specify `nil` to ignore the error information.
+
+ @return `YES` if the string was scheduled to send, otherwise - `NO`.
+ */
+- (BOOL)sendWithoutCopyingData:(nullable NSData *)data error:(NSError **)error NS_SWIFT_NAME(sendWithoutCopying(data:));
```
nlutsenko: Let's name it Swiftier and closer to Foundation in Swift 3.0 as well.
```
- (BOOL)sendDataNoCopy:(nullable NSData *)data error:(NSError **)error NS_SWIFT_NAME(send(dataNoCopy:));
```

nlutsenko:
Let's talk about the `delegate` method for copying data.

After looking deeper into the calls to `handleFrameWithData:opCode:` - there are numerous cases, where reading from that data would be incredibly unsafe, since we mutate the data and `NSData` is not thread-safe.I can see the value from not copying the incoming data, but at the same time using un thread-safe data can lead to crazy crazy problems on the receiving end and at the current moment.

Let's start with just the sending end of it and make this PR just about that, so we are unblocked on this PR.

For non-copying the frame data - looks like we can actually not copy the frame data, unless it's a web socket control frame, since we are going to simply re-create the `currentFrameData` again, so I bet there is a flow that we can figure out to be smart about copying, without requiring to implement the delegate method or configuration change.

nlutsenko:
Awe. Are you going to send a new one?

erikprice:
Yup, that PR was targeting your old `no-errors` branch. New one going to target `master`.

PR #429, opened 2016-06-29
Pluggable, more flexible, security policies.

michaelkirk:
### Motiviation

Extract @fredericjacobs' CertificateVerifier concept with @nlutsenko's
SRSecurityOptions into a pluggable SRSecurityPolicy protocol

This retains existing SSL configuration code paths, while allowing users
more flexibility to specify their own security policy. It's intended
that you're able to subclass an `AFSecurityPolicy` that conforms to the
`<SRSecurityPolicy>` protocol to share policy between websocket and
conventional requests.

Inspired by original "Require TLS 1.2 & enable pinning" pull request by
Frederic Jacobs (@fredericjacobs) at: https://github.com/facebook/SocketRocket/pull/274/files

### TODO

There are a couple things in particular that were kind of quick hacks, but wanted to get some feedback on the approach:

- [ ] Testing
- [X] I added a synchronized block in didConnect. Maybe we can refactor so this isn't reached at so many different points
- [x] Should SecurityPolicyBuilder be public?
- [X] Are you OK with defaulting to TLS 1.2 *only*?
- [X] ~~Implement no-encryption debug~~


facebook-github-bot-5:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-3:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

nlutsenko:
Love this PR, I think this is definitely the best of both worlds combined into one, with some polish required.

Putting few answers here the TODO questions you had:

>  Should SecurityPolicyBuilder be public?
Absolutely! I think we can structure it slightly differently, but creating different policies and using them should be public.

> Are you OK with defaulting to TLS 1.2 only?
Yes, as long as we make sure we are no restricting only to it, meaning people can downgrade security to SSLv3 or TLS 1.1.

SocketRocket/SRSecurityPolicy.h
```diff
@@ -0,0 +1,25 @@
+//
+// Copyright 2012 Square Inc.
```
nlutsenko: Nit: Since this is new code added and is not part of the original codebase - let's use the copyright statement from `SRSecurityPolicyBuilder`.

SocketRocket/SRSecurityPolicy.h
```diff
@@ -0,0 +1,25 @@
...
+// of patent rights can be found in the PATENTS file in the same directory.
+//
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol SRSecurityPolicy <NSObject>
+
+- (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust
```
nlutsenko: Nit: Without import for Security.framework - looks like importing this header will break compilation.
TL;DR Add `#import <Security/Security.h>` to the imports of this header.

SocketRocket/Internal/Security/SRSecurityPolicyBuilder.m
```diff
@@ -0,0 +1,213 @@
...
+    [stream setProperty:sslOptions forKey:(__bridge NSString *)kCFStreamPropertySSLSettings];
+}
+
+- (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust
+                  forDomain:(NSString*)domain
+{
+    SRDebugLog(@"Pinned cert count: %d", self.pinnedCertificates.count);
+    NSUInteger requiredCertCount = self.pinnedCertificates.count;
+    if (requiredCertCount == 0) {
+        @throw [NSException exceptionWithName:@"Evaluating security policy failed."
```
nlutsenko: Don't think this is required, since we validate the number of pinned certificates on creation of the policy.

SocketRocket/Internal/Security/SRSecurityPolicyBuilder.m
```diff
@@ -0,0 +1,213 @@
...
++ (SRWithoutSSLValidationSecurityPolicy *)withoutSSLValidationPolicy
+{
+    return [SRWithoutSSLValidationSecurityPolicy new];
+}
+
++ (id<SRSecurityPolicy>)createWithChainValidationEnabled:(BOOL)allowsUntrustedSSLCertificates
+                                      pinnedCertificates:(nullable NSArray *)pinnedCertificates
+{
+    if (pinnedCertificates)
+    {
```
nlutsenko: Nit: Please use common code style that we have aka braces are on the same line as `if` or `else`.

SocketRocket/Internal/Security/SRSecurityPolicyBuilder.m
```diff
@@ -0,0 +1,213 @@
...
+
+@implementation SRWithoutSSLValidationSecurityPolicy
+
+- (void)updateSecurityOptionsInStream:(NSStream *)stream
+{
+    // Enforce TLS 1.2
+    [stream setProperty:(__bridge id)CFSTR("kCFStreamSocketSecurityLevelTLSv1_2") forKey:(__bridge id)kCFStreamPropertySocketSecurityLevel];
+
+    // Do not validate certificate chain for the stream.
+    NSDictionary<NSString *, id> *sslOptions = @{ (__bridge NSString *)kCFStreamSSLValidatesCertificateChain : @(NO) };
```
nlutsenko: Nit: `@NO` :)

SocketRocket/SRWebSocket.m
```diff
@@ -427,33 +438,39 @@ - (void)_readHTTPHeader;
 
 - (void)didConnect;
 {
-    SRDebugLog(@"Connected");
+    // Potentially called through multiple code paths.
+    // Make sure we only respond once.
+    @synchronized(self) {
```
nlutsenko: You can avoid re-structuring all of this by simply adding `_requestRequiresSecurity` boolean variable and assigning it to `SRURLRequiresSSL()` to it.

nlutsenko:
Few initial comments and nits. The most crucial one is around `requestRequiresSSL` and changes around it.

Also, around `SRSecurityPolicyBuilder` I think it's closer to Cocoa standards to use class cluster pattern here instead of Factory one. A good example is `NSURLSessionConfiguration`, where it has class constructors for creating different configurations or you can subclass it or you can create one and set it up.

So here, we would do `SRSecurityPolicy` as a class with the same methods as the protocol, as well as class constructors for other policy subclasses that are internal, but control the behavior.

@michaelkirk, let me know your thoughts about all of this, and whether it makes sense.

michaelkirk:
> So here, we would do SRSecurityPolicy as a class with the same methods as the protocol, as well as class constructors for other policy subclasses that are internal, but control the behavior.

One concern we are trying to address is sharing a security policy from AFNetworking. It probably is more straight forward to implement a SRSecurityPolicy class like you've said, and then just delegate `evaluateTrustForDomain:WithTrust` to our AFSecurityPolicy.

michaelkirk:
OK, I've refactored the Factory/Protocol into a couple of concrete classes. Take a peak and let me know what you think.

I'm leaving my commits separate for the sake of a PR dialogue, but would suggest smashing them when merging.

michaelkirk:
Here's an example of how we're using this to delegate domain verification to our existing AFNetworking security policy: https://github.com/WhisperSystems/SignalServiceKit/pull/15

SocketRocket/Internal/Security/SRPinningSecurityPolicy.m
```diff
@@ -0,0 +1,68 @@
...
+
+#import <Foundation/Foundation.h>
+
+#import "SRLog.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SRPinningSecurityPolicy ()
+
+@property (nonatomic, readonly) NSArray *pinnedCertificates;
```
nlutsenko: `nonatomic, copy, readonly`

SocketRocket/SRSecurityPolicy.h
```diff
@@ -0,0 +1,67 @@
...
+//
+
+#import <Foundation/Foundation.h>
+#import <Security/Security.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SRSecurityPolicy : NSObject
+
+@property (nonatomic) BOOL validatesCertificateChain;
```
nlutsenko: `nonatomic, assign`, also we probably should make it `readonly`.

SocketRocket/SRSecurityPolicy.h
```diff
@@ -0,0 +1,67 @@
...
++ (instancetype)pinnningPolicyWithCertificates:(NSArray *)pinnedCertificates;
+
+/**
+ Specifies socket security and optional certificate chain validation.
+ 
+ @param isEnabled Whether or not to validate the SSL certificate chain. If you
+ are considering using this method because your certificate was not issued by a
+ recognized certificate authority, consider using `pinningPolicyWithCertificates` instead.
+ */
+- (instancetype)initWithValidatesCertificateChain:(BOOL)isEnabled NS_DESIGNATED_INITIALIZER;
```
nlutsenko: What do you think of a name `initWithCertificationChainValidationEnabled` and accordingly renaming the property?
The current one sounds a little bit weird.

SocketRocket/SRSecurityPolicy.h
```diff
@@ -0,0 +1,67 @@
...
+ you're free to override it in a subclass. See `SRPinningSecurityPolicy.h` for 
+ an example.
+
+ @param serverTrust The X.509 certificate trust of the server.
+ @param domain The domain of serverTrust.
+
+ @return Whether or not to trust the server.
+
+ */
+- (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust
```
nlutsenko: Nit: Make it one line, since we can easily fit it, as well as please add a space in `NSString *`

SocketRocket/SRWebSocket.h
```diff
@@ -11,6 +11,8 @@
 
 #import <Foundation/Foundation.h>
 
+#import "SRSecurityPolicy.h"
```
nlutsenko: Let's class forward here, instead of adding an explicit import.

nlutsenko:
Almost there... Few nits and renames. and this is good to go.
Also, feel free to rebase and squash the whole thing into a single commit (I am usually reading all of it anyway).

SocketRocket/SRSecurityPolicy.m
```diff
@@ -0,0 +1,76 @@
...
+    // Enforce TLS 1.2
+    [stream setProperty:(__bridge id)CFSTR("kCFStreamSocketSecurityLevelTLSv1_2") forKey:(__bridge id)kCFStreamPropertySocketSecurityLevel];
+
+    // Validate certificate chain for this stream if enabled.
+    NSDictionary<NSString *, id> *sslOptions = @{ (__bridge NSString *)kCFStreamSSLValidatesCertificateChain : @(self.certificateChainValidationEnabled) };
+    [stream setProperty:sslOptions forKey:(__bridge NSString *)kCFStreamPropertySSLSettings];
+}
+
+- (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust
+                  forDomain:(NSString*)domain
```
nlutsenko: Nit: Missed this one (move to the same line and add space between `NSString` and `*`.

SocketRocket/SRSecurityPolicy.m
```diff
@@ -0,0 +1,76 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SRSecurityPolicy ()
+
+@property (nonatomic, readonly) BOOL certificateChainValidationEnabled;
+
+@end
+
+/**
```
nlutsenko: This probably needs to live in the header instead of an implementation.

SocketRocket/SRSecurityPolicy.m
```diff
@@ -0,0 +1,76 @@
...
+#import "SRSecurityPolicy.h"
+#import "SRPinningSecurityPolicy.h"
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SRSecurityPolicy ()
+
+@property (nonatomic, readonly) BOOL certificateChainValidationEnabled;
```
nlutsenko: Nit: `nonatomic, assign, readonly`.

SocketRocket/SRSecurityPolicy.m
```diff
@@ -0,0 +1,76 @@
...
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree. An additional grant
+// of patent rights can be found in the PATENTS file in the same directory.
+//
+
+#import "SRSecurityPolicy.h"
+#import "SRPinningSecurityPolicy.h"
+
+#import <Foundation/Foundation.h>
```
nlutsenko: Nit: No need for import here, since we have one in the header file.

SocketRocket/SRSecurityPolicy.h
```diff
@@ -0,0 +1,65 @@
...
+ you're free to override it in a subclass. See `SRPinningSecurityPolicy.h` for
+ an example.
+
+ @param serverTrust The X.509 certificate trust of the server.
+ @param domain The domain of serverTrust.
+
+ @return Whether or not to trust the server.
+
+ */
+- (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust forDomain:(NSString*)domain;
```
nlutsenko: Nit: Add space between `NSString` and `*`.

SocketRocket/SRSecurityPolicy.h
```diff
@@ -0,0 +1,65 @@
...
++ (instancetype)pinnningPolicyWithCertificates:(NSArray *)pinnedCertificates;
+
+/**
+ Specifies socket security and optional certificate chain validation.
+
+ @param isEnabled Whether or not to validate the SSL certificate chain. If you
+ are considering using this method because your certificate was not issued by a
+ recognized certificate authority, consider using `pinningPolicyWithCertificates` instead.
+ */
+- (instancetype)initWithCertificateChainValidationEnabled:(BOOL)isEnabled NS_DESIGNATED_INITIALIZER;
```
nlutsenko: Nit: `enabled` instead of `isEnabled`, since the latter is usually a getter, not a property.

nlutsenko:
Sorry, another set of nits that I was able to spot.
I can merge this as is and fix it later, or feel free to update it (we are waiting on Travis anyway).

michaelkirk:
> Sorry, another set of nits that I was able to spot.

No worries. This is maybe the greatest code review I've ever had. =)

SocketRocket/SRSecurityPolicy.h
```diff
@@ -0,0 +1,68 @@
...
+ you're free to override it in a subclass. See `SRPinningSecurityPolicy.h` for
+ an example.
+
+ @param serverTrust The X.509 certificate trust of the server.
+ @param domain The domain of serverTrust.
+
+ @return Whether or not to trust the server.
+
+ */
+- (BOOL)evaluateServerTrust:(SecTrustRef)serverTrust forDomain:(nullable NSString *)domain;
```
nlutsenko: Should this really be `nullable`? Looks like our usage of it is to validate the host of the URL, which is required to be there for SRWebSocket to work.

SocketRocket/SRSecurityPolicy.h
```diff
@@ -0,0 +1,68 @@
...
+ This method should be used when responding to an authentication challenge from
+ a server. In the default implemenation, no further validation is done here, but
+ you're free to override it in a subclass. See `SRPinningSecurityPolicy.h` for
+ an example.
+
+ @param serverTrust The X.509 certificate trust of the server.
+ @param domain The domain of serverTrust.
+
+ @return Whether or not to trust the server.
+
```
nlutsenko: SuperNit: Remove empty line.

nlutsenko:
A single question about nullability there... I suspect you are getting a warning there?

SocketRocket/SRSecurityPolicy.h
michaelkirk: Hrm... I wasn't getting a warning, but it occured to me that  `NSURL -host` is nullable. 

https://github.com/facebook/SocketRocket/pull/429/files#diff-23371af810c948f5b65a136238d482e7R1452

nlutsenko:
The thing about nullability there is that if the `domain` is null - everything is f**ed at that point, meaning that we probably shouldn't force people when implementing subclasses of `SRSecurityPolicy` to accept a `nullable` domain.

michaelkirk:
> we probably shouldn't force people when implementing subclasses of SRSecurityPolicy to accept a nullable domain.

Fine by me - Just pushed up that change. 

Feel free to make any stylistic changes from here, but ping me if you have questions. Will be happy to have this merged. =)

nlutsenko:
This actually looks perfect now!
I am going to merge this in once Travis passes on it.

nlutsenko:
Thank you so much for making this happen!

michaelkirk:
Ha. I wonder how many other people's CI failed due to today's Rails 5 release. :\

michaelkirk:
Can you restart the build after #430 (or something similar) is merged?

nlutsenko:
I believe this needs rebase actually, since it's going to test your specific branch and not the commit applied on top of master.

nlutsenko:
This indeed fails tests, I am investigating right now.

nlutsenko:
I bet it's due to the fact that we are still trying to validate the security policy, even though it's not required.

SocketRocket/SRWebSocket.m
```diff
@@ -449,11 +471,8 @@ - (void)didConnect;
 - (void)_updateSecureStreamOptions
 {
     SRDebugLog(@"Setting up security for streams.");
-    [_securityOptions updateSecurityOptionsInStream:_inputStream];
-    [_securityOptions updateSecurityOptionsInStream:_outputStream];
-
-    SRDebugLog(@"Allows connection any root cert: %d", _securityOptions.validatesCertificateChain);
-    SRDebugLog(@"Pinned cert count: %d", _securityOptions.pinnedCertificates.count);
+    [_securityPolicy updateSecurityOptionsInStream:_inputStream];
```
nlutsenko: Failing tests and probably is going to fail to connect to `http` hosts with this.
Please wrap 474-475 in a `if (_requestRequiresSSL)` check.

nlutsenko:
Should be all good now, let's wait for Travis 😁

michaelkirk:
Great. Thanks for your prompt attention @nlutsenko! 🙏

PR #430, opened 2016-06-30
specify ruby version for CI

michaelkirk:
CI is currently failing because the default ruby on travis.org is 2.0, but a dependency xcpretty, requires the latest ActiveSupport. Version 5 of Active Support was just released today, which requires ruby >=2.2

I'm only 51% sure that this will work, but let's see what happens. I couldn't find documentation on specifying ruby versions for non-ruby projects.


nlutsenko:
It's actually cocoapods-core, not `xcpretty`.
This worked! Great! Thank you!
Merging now...

PR #431, opened 2016-07-01
Fix errors in TestChat application.

nlutsenko:
This was using an old API that we didn't release, but never got updated to the new one.

### DIFF
--- a/TestChat/TCViewController.m
+++ b/TestChat/TCViewController.m
@@ -99,7 +99,7 @@
 
 - (void)sendPing:(id)sender;
 {
-    [_webSocket sendPing:nil];
+    [_webSocket sendPing:nil error:NULL];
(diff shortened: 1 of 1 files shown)


TestChat/TCViewController.m
```diff
@@ -99,7 +99,7 @@ - (IBAction)reconnect:(id)sender
 
 - (void)sendPing:(id)sender;
 {
-    [_webSocket sendPing:nil];
+    [_webSocket sendPing:nil error:nil];
```
richardjrossiii: Shouldn't this be `NULL`, and not `nil`? Pointer to pointer vs pointer, etc.

nlutsenko: (－‸ლ)

PR #437, opened 2016-07-06
Update all targets to use shared configuration files from xctoolchain.

nlutsenko:


SocketRocket/SRWebSocket.m
```diff
@@ -560,7 +554,7 @@ - (void)closeWithCode:(NSInteger)code reason:(NSString *)reason;
         NSMutableData *mutablePayload = [[NSMutableData alloc] initWithLength:sizeof(uint16_t) + maxMsgSize];
         NSData *payload = mutablePayload;
 
-        ((uint16_t *)mutablePayload.mutableBytes)[0] = EndianU16_BtoN(code);
+        ((uint16_t *)mutablePayload.mutableBytes)[0] = CFSwapInt16BigToHost((uint16_t)code);
```
richardjrossiii: You could always use the POSIX defined `ntohs` (network-to-host-short), `ntohl` (network-to-host-long), and friends....

nlutsenko:
Passes everything, also I re-enabled sanitize flags and fixed undefined behavior.

PR #440, opened 2016-07-08
Make payload masking use vector instructions.

richardjrossiii:
This is up to 50x faster when running on an ARM64 device, and effects
every payload we send out from the device.

### DIFF
--- a/SocketRocket/Internal/Utilities/SRSIMDHelpers.h
+++ b/SocketRocket/Internal/Utilities/SRSIMDHelpers.h
@@ -1,0 +1,19 @@
+//
+// Copyright (c) 2016-present, Facebook, Inc.
+// All rights reserved.
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree. An additional grant
+// of patent rights can be found in the PATENTS file in the same directory.
+//
+
+#import <Foundation/Foundation.h>
+
+/**
+ Unmask bytes using XOR via SIMD.
+
+ @param bytes    The bytes to unmask.
+ @param length   The number of bytes to unmask.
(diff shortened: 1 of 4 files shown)


SocketRocket/Internal/Utilities/SRSIMDHelpers.h
```diff
@@ -0,0 +1,21 @@
...
+/**
+ Unmask bytes using XOR via SIMD.
+
+ @param bytes    The bytes to unmask.
+ @param length   The number of bytes to unmask.
+ @param mask_key The mask to XOR with MUST be of length sizeof(uint32_t).
+ */
+void SRMaskBytesSIMD(uint8_t *bytes, size_t length, uint8_t *mask_key);
+
+void SRMaskBytesManual(uint8_t *bytes, size_t length, uint8_t *mask_key);
```
nlutsenko: Probably don't want to expose it.

SocketRocket/Internal/Utilities/SRSIMDHelpers.h
```diff
@@ -0,0 +1,21 @@
...
+#import <Foundation/Foundation.h>
+
+/**
+ Unmask bytes using XOR via SIMD.
+
+ @param bytes    The bytes to unmask.
+ @param length   The number of bytes to unmask.
+ @param mask_key The mask to XOR with MUST be of length sizeof(uint32_t).
+ */
+void SRMaskBytesSIMD(uint8_t *bytes, size_t length, uint8_t *mask_key);
```
nlutsenko: Please use camel-case instead of snake case.

nlutsenko:
Please fix nits before merging.

PR #449, opened 2016-07-28
Unify usage of default buffer size between SRWebSocket and SRProxyConnect.

nlutsenko:


SocketRocket/Internal/SRConstants.m
```diff
@@ -0,0 +1,19 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree. An additional grant
+// of patent rights can be found in the PATENTS file in the same directory.
+//
+
+#import "SRConstants.h"
+
+size_t SRDefaultBufferSize(void) {
+    static size_t size;
+    static dispatch_once_t onceToken;
```
richardjrossiii: If we really cared, this could be done without a dispatch_once. We have a guarantee (at least on all arch that we support) that all loads/stores of WORD size are guaranteed to be atomic, so because the initializer is so lightweight (`getpagesize()` should just be a single instruction on our targets), as an uber-performance improvement, you could simply do the following:

```
static size_t size = 0;
if (size == 0) {
    size = getpagesize();
}
```

This works because of two things -
`size_t` is never greater than the word size of the memory bus.
`getpagesize()` will always return the same value, and super fast.

Now, honestly? I don't think this matters. Just something super small I noticed.

nlutsenko: This is good... Care to send a PR that fixes it?

richardjrossiii: Actually, in this case it's not an optimization.

On FreeBSD libc (or, apples fork of it: http://opensource.apple.com/source/Libc/Libc-1082.50.1/gen/FreeBSD/getpagesize.c), during the initial call to `getpagesize()`, they always make a syscall until they can cache the result. 

Now, AFIAK, `dispatch_once`, as it operates purely in userland would be faster than doing two syscalls required for `getpagesize()`.

(But note that libc does its own caching, which may just make it even better to have a direct call through to `getpagesize()` without a cache here).

PR #453, opened 2016-07-30
Update configurations and unbreak compilation of a sample app.

### LINKED ISSUE #454: 'UIKit/UIKit.h' file not found
Open project SocketRocket.xcodeproj, just support for run on mac, don't support iphone, but the source code is for ios, so how to solve the question?

nlutsenko:
We moved to shared configurations, but never got to updating the test chat app.

Fixes #454


SocketRocket/Internal/Utilities/SRURLUtilities.m
```diff
@@ -55,7 +55,15 @@ extern BOOL SRURLRequiresSSL(NSURL *url)
             break;
         case NSURLNetworkServiceTypeVoIP: {
             networkServiceType = NSStreamNetworkServiceTypeVoIP;
-            if (NSFoundationVersionNumber > NSFoundationVersionNumber_iOS_8_3) {
+
```
richardjrossiii: This completely changes the logic of the check here, near as I can tell. We probably shouldn't ever compare to `NSFoundationVersionNumber10_10_Max`, or `DBL_MAX` here. The entire check is to determine if `PushKit` is available, which can be done in several other ways than just a version check if we want.

The other choice being, of course, to just kill the print statement altogether, if we don't think it accomplishes anything.

nlutsenko: Wait, how is this changing the logic? I don't think I understand.

richardjrossiii:
Also: are the changes to `SRURLUtilities` relevant to this PR? I don't want to hold up unbreaking of the sample app because of some other minor changes.

nlutsenko:
Removed the print statement all together at this point.

Issue #421, opened 2016-06-23
-[SRProxyConnect _dequeueInput] crash


codefei:
the master branch, pod "SocketRocket", :git => 'https://github.com/facebook/SocketRocket.git', :commit => '0af058c'

<img width="560" alt="2016-06-23 5 52 10" src="https://cloud.githubusercontent.com/assets/7637224/16299116/34c43ad6-396b-11e6-8566-f2355fb8f1c9.png">

codefei:
_inputQueue may be accessed by different threads, it's not thread safe?

nlutsenko:
@codefei, can you share the full stack trace for all the threads by any chance?
Would be helpful to see if any other thread is trying to access the same thing.

nlutsenko:
`inputQueue` is ok to be not thread-safe, since it's accessed from a context of a single thread aka `SRRunLoopThread`, that's why I am wondering on the state of other threads...

codefei:
here is the full stack, and I have checked other thread state doesn't find useful info.

[stack.txt](https://github.com/facebook/SocketRocket/files/331248/stack.txt)

michaelkirk:
Seeing a similar crash. 

Crash report for all threads: https://gist.github.com/michaelkirk/8ceccf89d270c2df37fdc9d28a2d1e40 

It wasn't in our top crashes until updating from 7373546fba1852c1b946751f1af227de6bbb6036 to b0aabde7055c81bff71aa9c835f1d69eca923d03.

This could be because it's a new crash, or because we've substantially reduced the noise of our other top crashes in the meanwhile (substantially because of https://github.com/facebook/SocketRocket/issues/156 - thanks!).

Note, we're running a modified version to support our pinning policy - so actually the exact upgrade was from https://github.com/FredericJacobs/SocketRocket/commit/a062cc7708b78d7673776140f0e1843906862343 to https://github.com/WhisperSystems/SocketRocket/commit/587ad297eb63eb0d64d4caeb32a7da646ad1132b.

erikprice:
We have a related crash – it's in the same method, but a different error message:

**Fatal Exception: NSInternalInconsistencyException<br />
-[__NSCFArray removeObjectAtIndex:]: mutating method sent to immutable object**

```
Fatal Exception: NSInternalInconsistencyException
0  CoreFoundation                 0x2098b91b __exceptionPreprocess
1  libobjc.A.dylib                0x20126e17 objc_exception_throw
[...]

PR #455, opened 2016-08-04
Optimize input queue processing in SRProxyConnect.

nlutsenko:
Purely to tackle the theory described in https://github.com/facebook/SocketRocket/issues/421#issuecomment-237535065

### DIFF
--- a/SocketRocket/Internal/Proxy/SRProxyConnect.m
+++ b/SocketRocket/Internal/Proxy/SRProxyConnect.m
@@ -407,13 +407,17 @@
 - (void)_dequeueInput
 {
     while (_inputQueue.count > 0) {
-        NSData *data = _inputQueue[0];
-        [self _proxyProcessHTTPResponseWithData:data];
+        NSData *data = _inputQueue.firstObject;
         [_inputQueue removeObjectAtIndex:0];
+
+        // No need to process any data further, we got the full header data.
+        if ([self _proxyProcessHTTPResponseWithData:data]) {
+            break;
+        }
     }
 }
 //handle checking the proxy  connection status
-- (void)_proxyProcessHTTPResponseWithData:(NSData *)data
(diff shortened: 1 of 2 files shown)


erikprice:
This looks good, @nlutsenko. Thanks for submitting this.

We can't reproduce the bug that this fixes, but we've seen it in the field. We are still using a fork of SocketRocket, if you remember. I'll apply this commit to our fork and it will go out in our next App Store release, and we'll let you know if we see a drop in crash reports for this problem.

richardjrossiii:
What does the stack trace look like when `-dequeueInput` is invoked? Are there still ivar references to the SRProxyConnect around that aren't strong at that time?

Honestly, I think the proper solution here is to either

- A: Hold a strong reference to self for the duration of `deuqueInput`.
- B: Ensure that the ProxyConnect object is not deallocated in the completion handler.

Having this strange scenario where you potentially cannot use any of the variables in `self` after calling `_proxyProcessHTTPResponseWithData` seems dangerous, and likely to break again in the future.

erikprice:
@richardjrossiii said

> What does the stack trace look like when -dequeueInput is invoked?

You can see the stack trace from two separate apps that have encountered this problem in #421. I'm not sure if there is an ivar reference to the SRProxyConnect within the SocketRocket project, but it does seem possible that the SRProxyConnect object is still set as the delegate of a NSStream object at that time, and this property is listed as having `assign` semantics [in Apple's header file](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSStream_Class/#//apple_ref/occ/instm/NSStream/delegate).

Does that help with your assessment?

nlutsenko:
Per our discussion - I ended up just moving the cleanup for proxy connect to the work queue, instead of creating an autoreleasing local scope variable. Biggest reason - it's disconnected and most of the calls are in a separate method.

richardjrossiii:
#shipit

PR #469, opened 2016-09-13
Fix memory leak by freeing mutex.

woudini:
Instruments shows a leak of category Malloc 64 Bytes from `SRMutexInitRecursive` after re-initializing a `SRWebSocket` property.

I have an app that uses a `SRWebSocket` property for a few different subscriptions that need to be active at different times. As a temporary solution I am manually closing the socket to unsubscribe from everything and re-initializing the socket since I can only call `open` on each instance once.

After calling `free` on mutex, the memory leak no longer appears in Instruments.


nlutsenko:
Hey @woudini, thanks for the contribution!
This is an amazing catch! You are absolutely right, there is a need to free the mutex after it is destroyed. Will merge in after the tests pass.

woudini:
Awesome, thanks @nlutsenko for the quick reply!

PR #501, opened 2017-02-06
Adding bearer authentication support

tmughal:
This commit adds support for bearer authentication during the WebSocket handshake request.


gaming-hacker:
@ tmughal  nice patch.  to stop the compiler from complaining that the self could be nil. i also changed this delegate to a strong reference

`@property (nonatomic, strong) id <SRWebSocketAuthenticationDelegate> authenticationDelegate;`

tmughal:
@gaming-hacker Glad it was useful

PR #503, opened 2017-02-19
Fixes debug logging

feosuna1:
It looks like `SRDebugLog` hasn't been used a in a while because uncommenting `#define SR_DEBUG_LOG_ENABLED` caused a bunch of compiler errors.

This change fixes up the use of `SRDebugLog` so that uncommenting `#define SR_DEBUG_LOG_ENABLED` will properly emit log statements.

Additionally, I added the `NS_FORMAT` attribute to `SRErrorLog`, which hints to the compiler that `SRErrorLog` has a format string parameter.  This caused a few analyzer warnings that this PR also fixes.

### DIFF
--- a/SocketRocket/Internal/Proxy/SRProxyConnect.m
+++ b/SocketRocket/Internal/Proxy/SRProxyConnect.m
@@ -337,7 +337,7 @@
 
 - (void)stream:(NSStream *)aStream handleEvent:(NSStreamEvent)eventCode;
 {
-    SRDebugLog(@"stream handleEvent %u", eventCode);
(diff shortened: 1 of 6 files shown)


feosuna1:
I'm not sure why 4c62964 generated a test failure.  I tested it locally and I had no failures. I'm not what the error was, as the error log doesn't show the error message.  I added 70cc4b4 which appends the error message to the assertion.

feosuna1:
Travis [failed](https://travis-ci.org/facebook/SocketRocket/jobs/203189461#L920) this time because it timed out while building.

PR #504, opened 2017-02-20
Fix memory leaks

### LINKED ISSUE #499: Ever increasing Memory usage on reconnect
I am running [this](https://github.com/facebook/SocketRocket/blob/41b57bb2fc292a814f758441a05243eb38457027/TestChat/TCViewController.m#L88) every two seconds if there is no connection to the socket:

```
- (void)openSocket
{
    serverReplied = false;
    streamOpen = false;
    
    _webSocket.delegate = nil;
    [_webSocket close];
    
    _webSocket = [[SRWebSocket alloc] initWithURL:[NSURL URLWithString:@"wss://s.notifi.it"]];
    _webSocket.delegate = (id)self;
    
    [_webSocket open];
}
```
but this seems to not be freeing up memory after it is called. 

<sub>I am a massive amateur to freeing up memory in obj-c and am finding it a little confusing atm.</sub>

feosuna1:
I believe this PR fixes #499 and addresses a few more memory related issues I found.  

The root cause of #499 was that the timeout timer was strongly retaining the `SRWebSocket` instance and wouldn't release it until after the timer popped.

I did find a bonafide memory leak, if you attempted to close the `SRWebSocket` before `SRProxy` had a chance to return, the the `SRWebSocket` would never clean up its strong reference to itself and the object would be trapped in memory forever.

### DIFF
--- a/SocketRocket/SRWebSocket.h
+++ b/SocketRocket/SRWebSocket.h
@@ -85,7 +85,7 @@ extern NSString *const SRHTTPResponseErrorKey;
 /**
  An operation queue for scheduling the delegate calls.
 
- If `nil` and `delegateOperationQueue` is `nil`, the socket uses main queue for performing all delegate method calls.
+ If `nil` and `delegateDispatchQueue` is `nil`, the socket uses main queue for performing all delegate method calls.
  */
 @property (nullable, nonatomic, strong) NSOperationQueue *delegateOperationQueue;
 
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -305,23 +305,33 @@ NSString *const SRHTTPResponseErrorKey = @"HTTPResponseStatusCode";
     NSAssert(self.readyState == SR_CONNECTING, @"Cannot call -(void)open on SRWebSocket more than once.");
 
     _selfRetain = self;
+    __weak typeof(self) wself = self;
 
     if (_urlRequest.timeoutInterval > 0) {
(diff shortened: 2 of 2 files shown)


maxisme:
@feosuna1 Thanks for this! I did not notice this until now. I have a very brute force technique of setting up and pursuing my socket where I check for a `pong` after a `ping` (every 30 seconds) and if I do not receive the pong I close the socket. I then have every 2 seconds:
```
        if(!streamOpen){
            [self openSocket];
        }else if(!userAuthed){
            [self authUser];
        }
```

Is there a better way to do this?

feosuna1:
@maxisme, you should probably wait to reattempt a  connection until the previous attempt fails. Calling it bunch of times in rapid succession could cause to the memory leaks described in this PR. 

After this patch is applied, your technique would work but I would personally refactor your code so that subsequent attempts are not made until previous connnections have been completely cleaned up.

feosuna1:
@maxisme, also I wouldn't close the previous connection until I got confirmation that it was properly opened.

ajwhite:
This looks like a great improvement, and appears to be the root cause of https://github.com/intercom/intercom-ios/issues/361.

@feosuna1 what would it take to get this across the finish line? Are you waiting for a reviewer?

Via git blame, cc @nlutsenko @mikelikespie

feosuna1:
@ajwhite yep, just need a code review and address any feedback.

zhoubo2015:
continue happen memory leak.

PR #506, opened 2017-02-21
Fix for SRWebSocket dealloc method

fscheer:
Modified SRWebSocket dealloc method to close streams before
setting stream delegate to nil;


dumganhar:
@fscheer , may I ask what's the issue of current code? Why should close streams before setting the delegate to nil?

PR #522, opened 2017-07-06
Handle IPv6 host in Request Headers

madlymad:
In case of IPv6 wrap the ip in brackets for "Host" and "Origin" headers
Without the brackets server responds with 400-Bad Request.


dumganhar:
@nlutsenko, do you get time to review this pull request? Could it be merged?

madlymad:
For some reason Travis does not like this PR I created a question requesting some support from [Autobahn](https://github.com/crossbario/autobahn-testsuite/issues/75) in order to identify if the problem is on the tests or the code and why this happens but I have no response so far.

But  in my case this solved the problem.

PR #525, opened 2017-07-20
Memory leak fix and cancel connection timeout timer in scheduleCleanup.

dumganhar:
This Pull Request fixes two issues:
1. Uses weak SRWebSocket object in the block of `_readUntilHeaderCompleteWithCallback`. Since consumers has a `_completion` block which will retain SRWebSocket instance, if we uses strong self, cycle reference will occur.
To Reproduce this issue, just simply connect an invalid website, e.g. ws:://echo111.websocket111.org111 , set a breakpoint in `SRWebSocket:dealloc`, it will never go there. This bug was found by analyzing `Allocation List` in the `Instruments` tool of Xcode.

2. Should cancel connection timeout timer in scheduleCleanup method, otherwise, the `SRWebSocket:dealloc` method will be invoked only after the time (default: 60s) have come, which may cause memory increase.
Since `dispatch_after` could not be cancelled, use `performSelector:withObject:afterDelay` & `cancelPreviousPerformRequestsWithTarget:selector:object` instead.

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -295,6 +295,14 @@ NSString *const SRHTTPResponseErrorKey = @"HTTPResponseStatusCode";
     return NO;
 }
 
+-(void)_onTimeout
+{
+    if (self.readyState == SR_CONNECTING) {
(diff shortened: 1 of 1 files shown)


dumganhar:
I don't know why travis build was failed. :(

gaming-hacker:
@dumganhar i got a warning when patching this

`Weak variable 'wself' is accessed multiple times in this block but may be unpredictably set to nil`

PR #534, opened 2017-08-14
Deprecate SSL pinning and trust chain verification.

richardjrossiii:
Oh boy. Here's a controversial change.

Let's give a bit of backstory.

A few weeks ago, Facebook was contacted by a whitehat hacker (the good
guys) about a security vulnerability here in SocketRocket.

For those of you who are truly interested in what that security flaw
was, it is essentially the same flaw as outlined here:

https://www.synopsys.com/blogs/software-security/ineffective-certificate-pinning-implementations/

So, we were faced with a choice - quietly push out a patch, and hope
that eventually existing applications updated, or be transparent and
admit we screwed up.

This is us admititng we screwed up. And while yes, we could probably fix
the implementation. But we talked internally, and decided that the best
approach here is to completely remove the option for pinning.

For all of our existing users that use certificate pinning, while we
understand that in the past there has been a very large barrier to entry
with getting a CA to issue a certificate.

However, since the rollout of CAs like LetsEncrypt, there's become an
ever-dwindling reason to actually use self-signed or unsigned
certificates.

For this reason, we're going to go ahead and deprecate the APIs that
allow SSL pinning and disabling trust chain verification. The pinning
APIs are now going to throw an exception when invoked, and the trust
chain APIs have deprecation warnings.

If you are a user of these APIs, and you for some reason **CANNOT** use
a trust chain validated certificate, PLEASE speak up. While we cannot
think of any reason to use those kinds of certificates, it's entirely
possible we overlooked something. We'll leave this pullrequest unmerged
for a two week period (Monday, August 28th, 2017), at which point,
unless we have feedback convincing us otherwise, we will go ahead with
this change.

### DIFF
--- a/SocketRocket/Internal/Security/SRPinningSecurityPolicy.h
+++ b/SocketRocket/Internal/Security/SRPinningSecurityPolicy.h
@@ -13,6 +13,11 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/** 
+ * NOTE: While publicly, SocketRocket does not support configuring the security policy with pinned certificates,
+ * it is still possible to manually construct a security policy of this class. If you do this, note that you may
+ * be open to MitM attacks, and we will not support any issues you may have. Dive at your own risk.
+ */
 @interface SRPinningSecurityPolicy : SRSecurityPolicy
 
 - (instancetype)initWithCertificates:(NSArray *)pinnedCertificates;
--- a/SocketRocket/NSURLRequest+SRWebSocket.m
+++ b/SocketRocket/NSURLRequest+SRWebSocket.m
@@ -23,21 +23,18 @@ static NSString *const SRSSLPinnnedCertificatesKey = @"SocketRocket_SSLPinnedCer
 
 - (nullable NSArray *)SR_SSLPinnedCertificates
 {
-    return [NSURLProtocol propertyForKey:SRSSLPinnnedCertificatesKey inRequest:self];
+    return nil;
 }
 
 @end
 
 @implementation NSMutableURLRequest (SRWebSocket)
 
-- (nullable NSArray *)SR_SSLPinnedCertificates
-{
-    return [NSURLProtocol propertyForKey:SRSSLPinnnedCertificatesKey inRequest:self];
-}
-
 - (void)setSR_SSLPinnedCertificates:(nullable NSArray *)SR_SSLPinnedCertificates
 {
-    [NSURLProtocol setProperty:[SR_SSLPinnedCertificates copy] forKey:SRSSLPinnnedCertificatesKey inRequest:self];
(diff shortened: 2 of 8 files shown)


SocketRocket/Internal/Security/SRPinningSecurityPolicy.h
```diff
@@ -13,6 +13,9 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+// NOTE: While publicly, SocketRocket does not support configuring the security policy with pinned certificates,
```
nlutsenko: Nit: Use a section comment aka `/** ... */` as well as document the class itself?

SocketRocket/NSURLRequest+SRWebSocket.m
```diff
@@ -23,21 +23,18 @@ @implementation NSURLRequest (SRWebSocket)
...
 
-- (nullable NSArray *)SR_SSLPinnedCertificates
-{
-    return [NSURLProtocol propertyForKey:SRSSLPinnnedCertificatesKey inRequest:self];
-}
-
 - (void)setSR_SSLPinnedCertificates:(nullable NSArray *)SR_SSLPinnedCertificates
 {
-    [NSURLProtocol setProperty:[SR_SSLPinnedCertificates copy] forKey:SRSSLPinnnedCertificatesKey inRequest:self];
+    [NSException raise:NSGenericException
```
nlutsenko: Nit: Not sure if `Generic` or `Parameter` exception? You sure about `Generic`? Not that it matters much tho...

SocketRocket/Internal/Security/SRPinningSecurityPolicy.h
richardjrossiii: It is a private class as of right now, I don't think full documentation is required at this time. I will change to using /** comments though.

chax:
I know this is already merged, but here is an example where you would need self signed certificate and you don't have a possibility of using CA to make valid trusted certificate.

Imagine this, you have a device which has REST API and it's only accessible locally over LAN. Since it is local device, it has a local IP address, but doesn't have domain. You still want to use SSL for this REST API, that is, access it over HTTPS. Since it has only local IP address and no domain, it's impossible for CA to issue certificate and our only option is to use self signed certificate.

davidperrenoud:
In the wake of [Project Atlas](https://techcrunch.com/2019/01/29/facebook-project-atlas/) and the [lack of adoption of HPKP](https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/he9tr7p3rZ8/eNMwKPmUBAAJ), wouldn't certificate pinning be useful in preventing eavesdropping by using a rogue root certificate?

Or is there a way to enforce Certificate Transparency inside my iOS/Android app?

Issue #500, opened 2017-01-23
Error ITMS-90056: Info.plist file is missing the required key: CFBundleVersion


dstarke:
When built from master, the framework cannot be included in an app store submission. ITMS returns the error ITMS-90056 and reports that the Info.plist file is missing the `CFBundleVersion` key.

It looks like this may be caused by an earlier change that updated all the targets to use shared configurations. Prior to that, the build configurations set a `CURRENT_PROJECT_VERSION` build variable (it was set to `1`), and the Info.plist sets the `CFBundleVersion` to the value of that build variable. However, the .xcconfig files do not set `CURRENT_PROJECT_VERSION`, so the built framework's `CFBundleVersion` is empty.

There are two easy fixes: either update the configurations to set `CURRENT_PROJECT_VERSION`, or update the Info.plist so that `CFBundleVersion` is set to a valid value instead of `$(CURRENT_PROJECT_VERSION)`

Droppix:
+1

RickOHara:
+1

hatdragon:
+1

SjoerdPerfors:
+10

lswith:
+1

WFNeIn:
I set it but it doesn't work. it is still report same error

dzenbot:
++

sanojbluejeans:
+1 I am facing exact same issue. Did anybody resolve this issue?

dzenbot:
We are checking the commit in https://github.com/facebook/SocketRocket/pull/535 that fixes this issue.

sanojbluejeans:
@dzenbot I am pointing to master currently but still I am facing same issue. Is that fix is not yet merged?
ERROR ITMS-90056: "This bundle Payload/MyApp.app/Frameworks/SocketRocket.framework is invalid. The Info.plist file is missing the required key: CFBundleVersion."

dzenbot:
It's is a regression in `master`. You should check the commit from the PR I posted.

sanojbluejeans:
@dzenbot  That PR is open or are you referring any other one. please mention the commit id. Please let me know. I am blocked and cant push the app to app store.
https://github.com/facebook/SocketRocket/pull/535

sanojbluejeans:
@ I resolved it by adding the missing version created framework by Carthage.

PR #535, opened 2017-08-31
Fix for #500

lswith:
This fixes the issue #500


jimocallaghan:
Is there a chance this PR will get merged?  We're facing the same issue - need some of the other post r0.4.2 fixes also and without this are getting bounced on iTunes upload with ITMS-90056 - "The Info.plist file is missing the required key: CFBundleVersion."

iandundas:
Please can this be merged soon.

PR #559, opened 2018-02-26
Add specific bundle version

funkyboy:
When adding SocketRocket as a dependency via Carthage everything works fine, until you submit the app to the App Store. 
When you do you hit this validation error.

screen shot 2018-02-26 at 10 42 05

This PR fixes this error.


destryteeter:
Still relevant.

PR #610, opened 2019-04-11
Update CFBundleVersion plist

Shakarang:
Hello,
We are facing this issue when we try to upload our build on the App Store.
```
"This bundle Payload/MyApp.app/Frameworks/SocketRocket.framework is invalid. The Info.plist file is missing the required key: CFBundleVersion. Please find more information about CFBundleVersion at https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion"
```
The CFBundleVersion references a `CURRENT_PROJECT_VERSION`, which is not present in the project. So I've put the same version as in the short version.


nlutsenko:
Thanks.

PR #618, opened 2019-10-26
Update SRWebSocket.m

HarrysLapTimer:
Revised include statement (changed #import to #include) to have the proper include path used (C-source).

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -16,7 +16,7 @@
 #endif
 
 #ifdef HAS_ICU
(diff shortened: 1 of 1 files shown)


HarrysLapTimer:
CLA signed.

nlutsenko:
Since this is ObjC, not just raw C source file, the semantics of `#import` vs `#include` works the same way, and `#import` is generally preferred over #include, as it can do some caching (similar to what say `#pragma once` would do).

All-in-all, thank you for the contribution, but I don't think this is required.

Feel free to reopen and let me know if you think otherwise.

PR #687, opened 2024-08-27
Replace usage of deprecated APIs, fix implicit nullability conversion.

nlutsenko:
Closes #680

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -19,7 +19,7 @@
 #import <unicode/utf8.h>
 #endif
 
-#import <libkern/OSAtomic.h>
+#import <os/lock.h>
 
 #import "SRDelegateController.h"
 #import "SRIOConsumer.h"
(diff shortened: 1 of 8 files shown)


cipolleschi (approved):
Code looks good to me

SocketRocket/SRWebSocket.m
```diff
@@ -1421,7 +1426,17 @@ - (void)stream:(NSStream *)aStream handleEvent:(NSStreamEvent)eventCode
...
+            NSString *const host = _urlRequest.URL.host;
+            if (!host || host.length == 0) {
+                dispatch_async(_workQueue, ^{
+                    NSError *error = SRErrorWithDomainCodeDescription(NSURLErrorDomain,
+                                                                      NSURLErrorBadURL,
+                                                                      @"Unable to validate certificate for empty host.");
+                    [wself _failWithError:error];
+                });
+                return;
+            }
```
cipolleschi, line 1438: nit: why do we need to jump thread to emit an error?

nlutsenko: :) Great question, we actually don't, you are right!
Let me clean it up after, there are more cases like this, the worst case - it's an extra dispatch here in error case.

PR #693, opened 2024-10-07
Ensure readyState is set to SR_CLOSED on successful WebSocket closure

nnabeyang:
This PR fixes an issue where the `readyState` was not properly set to `SR_CLOSED` when the WebSocket closed successfully.

### Changes:
- Previously, during a successful WebSocket closure, the `readyState` remained at `SR_CLOSING` and was not updated to `SR_CLOSED`.
- With this fix, the `readyState` is now set to `SR_CLOSED` before the delegate method responsible for handling the WebSocket closure is called.

### DIFF
--- a/SocketRocket/SRWebSocket.m
+++ b/SocketRocket/SRWebSocket.m
@@ -1113,6 +1113,8 @@ static const uint8_t SRPayloadLenMask   = 0x7F;
             }
         }
 
+        self.readyState = SR_CLOSED;
+
         if (!_failed) {
             [self.delegateController performDelegateBlock:^(id<SRWebSocketDelegate>  _Nullable delegate, SRDelegateAvailableMethods availableMethods) {
                 if (availableMethods.didCloseWithCode) {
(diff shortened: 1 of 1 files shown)


SocketRocket/SRWebSocket.m
```diff
@@ -1113,6 +1113,8 @@ - (void)_pumpWriting
             }
         }
 
+        self.readyState = SR_CLOSED;
```
cipolleschi, line 1116: Shouldn't this code go inside the `if(!_failed)`?
What's the state of the WebSocket if we send a command to close it and it fails to close for whatever reason? Is it in closing state? is it open?

cipolleschi (changes requested):
Hi @nnabeyang, thanks for the contribution.
I think that the line you added is correct but not in the best place. I added a comment.

SocketRocket/SRWebSocket.m
nnabeyang: @cipolleschi Thank you! You're right, it does make more sense to put it inside `if(!_failed)`. I fixed it. 

https://github.com/facebookincubator/SocketRocket/pull/693/commits/2d81c50baf61b2ebbfa7a24c3c050abb3dde2092

PR #701, opened 2025-03-06
Support using TLS 1.3

saghul:
kCFStreamSocketSecurityLevelNegotiatedSSL is the only public non-deprecated constant which should negotiate connections with TLS >= 1.2.


saghul:
Ping @nlutsenko since I see you've been active last :-)

saghul:
Ping, anyone?

nicoabie:
Ping, we are blocked by this too

